PR #171, opened 2016-09-27
Terraform plugin with examples

chungers:
This PR includes a proof-of-concept integration using Terraform as an instance plugin.

The examples in the directory contain a full AWS vpc with instances.  With new instances are provisioned, terraform is used to apply and create the instance.  Describing the instances will parse the `.tfstate` file and search for instances with tags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ Several binaries are available. More detailed documentations can be found here
   + [`infrakit/cli`](./cmd/cli/README.md), the command line interface
   + [`infrakit/group`](./cmd/group/README.md), the default [group plugin](./spi/group)
   + [`infrakit/file`](./example/instance/file), an instance plugin using files
+  + [`infrakit/terraform`](./example/instance/terraform), an instance plugin integrating [Terraform](https://www.terraform.io)
   + [`infrakit/vagrant`](./example/instance/vagrant), an instance plugin using vagrant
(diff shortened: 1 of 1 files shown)


linsun:
What is the difference between instance-plugin-properties.json and instance-plugin-spec.json?  seems most of the properties are duplicated.

wfarner:
@linsun definitely non-obvious how these files differ.

`instance-plugin-properties.json` contains the JSON document that represents the terraform plugin's [`SpecPropertiesFormat`](https://github.com/docker/infrakit/blob/044c07a155d6ab893c1666de11bd68718ca57202/example/instance/terraform/plugin.go#L136-L139).

`instance-plugin-spec.json` contains the JSON document that represents [`instance.Spec`](https://github.com/docker/infrakit/blob/044c07a155d6ab893c1666de11bd68718ca57202/spi/instance/types.go#L24-L30), which nests the above document (as you pointed out).

The docs added in this PR describe how to simulate the instance plugin interface functions being called as they would be from the stock group plugin implementation, hence the two different structures.

@chungers do you think it would make sense to reorganize these docs to use groups instead of direct calls to the instance plugin?  I think that would more closely match how we expect users to interact with instance plugins.

chungers:
Yes the docs were written to show using the terraform plugin by itself but it's more illustrative to show this working with the group plugin. 

I will add a doc on that today. PR coming.

PR #192, opened 2016-10-04
More doc -- Terraform + InfraKit (demo script)

chungers:
+ This is the script for demonstrating InfraKit + Terraform integration, with updates to the README.
+ The exact scenario demo'd at the first InfraKit talk at LinuxCon Berlin 2016.  Included here so
people who are interested can try it out.


mnp:
I would be interested to see how the Terraform and Swarm examples interplay.

Our real stack uses both (TF makes instances, swarm deploys containers to them) and we are hoping InfraKit will tie them together for us.  A doc or working example would be nice.

PR #194, opened 2016-10-05
Added support for ENV var at the cli

FrenchBen:
Added support for env vars to define the Listen/tcp endpoint. This is especially useful when going through the tutorial and wanting to create the endpoint in say the `/tmp` dir, which does not require escalated privileges.

Introduces the env var `INFRAKIT_PLUGINS_DIR`.

Updated the tutorial to reflect some wording changes and the ability to point the listening socket to the `/tmp` dir of choice along with setting the plugin dir for the `cli` to use.

cc @chungers @wfarner

### DIFF
--- a/cmd/group/main.go
+++ b/cmd/group/main.go
@@ -39,9 +39,8 @@ var (
 func main() {
 
 	logLevel := len(log.AllLevels) - 2
-
-	listen := "unix:///run/infrakit/plugins/"
-	sock := "group.sock"
+	discoveryDir := "/run/infrakit/plugins/"
+	name := "group"
 
 	pollInterval := 10 * time.Second
 
@@ -61,20 +60,14 @@ func main() {
 				return nil
 			}
 
-			listen = viper.GetString("listen")
-			sock = viper.GetString("sock")
-
-			log.Infof("Parsing url: %s - %s\n", listen, sock)
+			discoveryDir = viper.GetString("discovery")
+			name = viper.GetString("name")
+			listen := fmt.Sprintf("unix://%s/%s.sock", path.Clean(discoveryDir), name)
 
 			// parse the listen string
 			listenURL, err := url.Parse(listen)
 			if err != nil {
 				return err
-			}
-
-			if listenURL.Scheme == "unix" {
-				log.Info("Unix scheme detected")
-				listenURL.Path = path.Join(listenURL.Path, sock)
 			}
 
 			log.Infoln("Starting discovery")
@@ -140,16 +133,12 @@ func main() {
 		},
 	})
 
-	log.Info("Adding cmd binds")
-	cmd.Flags().String("listen", listen, "listen address (unix or tcp) for the control endpoint")
-	log.Info("Viper Bind listen")
-	// Bind env var for plugin listen address
-	viper.BindEnv("listen", "INFRAKIT_PLUGINS_LISTEN")
+	cmd.Flags().String("discovery", discoveryDir, "Dir discovery path for plugin discovery")
 	// Bind Pflags for cmd passed
-	viper.BindPFlag("listen", cmd.Flags().Lookup("listen"))
-	cmd.Flags().String("sock", sock, "listen socket for the control endpoint")
-	// Bind Pflags for cmd passed
-	viper.BindPFlag("sock", cmd.Flags().Lookup("sock"))
+	viper.BindEnv("discovery", "INFRAKIT_PLUGINS_DIR")
+	viper.BindPFlag("discovery", cmd.Flags().Lookup("discovery"))
+	cmd.Flags().String("name", name, "listen socket name for the control endpoint")
+	viper.BindPFlag("name", cmd.Flags().Lookup("name"))
 	cmd.Flags().IntVar(&logLevel, "log", logLevel, "Logging level. 0 is least verbose. Max is 5")
 	cmd.Flags().DurationVar(&pollInterval, "poll-interval", pollInterval, "Group polling interval")
 
--- a/cmd/group/main.go
+++ b/cmd/group/main.go
@@ -84,11 +84,6 @@ func main() {
 				return flavor_client.PluginClient(callable), nil
 			}
 
-			log.Infoln("Starting plugin")
-
-			log.Infoln("Starting")
-			log.Infoln("Listening on:", listenURL.String())
-
 			_, stopped, err := util.StartServer(listenURL.String(), group_server.PluginServer(
 				group.NewGroupPlugin(
 					instancePluginLookup,
--- a/cmd/group/main.go
+++ b/cmd/group/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"path"
 	"path/filepath"
 	"time"
 
@@ -18,6 +19,7 @@ import (
 	instance_client "github.com/docker/infrakit/spi/http/instance"
 	"github.com/docker/infrakit/spi/instance"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 var (
@@ -38,7 +40,8 @@ func main() {
 
 	logLevel := len(log.AllLevels) - 2
 
-	listen := "unix:///run/infrakit/plugins/group.sock"
+	listen := "unix:///run/infrakit/plugins/"
+	sock := "group.sock"
 
 	pollInterval := 10 * time.Second
 
@@ -58,10 +61,20 @@ func main() {
 				return nil
 			}
 
+			listen = viper.GetString("listen")
+			sock = viper.GetString("sock")
+
+			log.Infof("Parsing url: %s - %s\n", listen, sock)
+
 			// parse the listen string
 			listenURL, err := url.Parse(listen)
 			if err != nil {
 				return err
+			}
+
+			if listenURL.Scheme == "unix" {
+				log.Info("Unix scheme detected")
+				listenURL.Path = path.Join(listenURL.Path, sock)
 			}
 
 			log.Infoln("Starting discovery")
@@ -90,9 +103,9 @@ func main() {
 			log.Infoln("Starting plugin")
 
 			log.Infoln("Starting")
-			log.Infoln("Listening on:", listen)
+			log.Infoln("Listening on:", listenURL.String())
 
-			_, stopped, err := util.StartServer(listen, group_server.PluginServer(
+			_, stopped, err := util.StartServer(listenURL.String(), group_server.PluginServer(
 				group.NewGroupPlugin(
 					instancePluginLookup,
 					flavorPluginLookup,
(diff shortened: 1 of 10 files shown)


cmd/group/main.go
```diff
@@ -38,7 +40,8 @@ func main() {
 
 	logLevel := len(log.AllLevels) - 2
 
-	listen := "unix:///run/infrakit/plugins/group.sock"
+	listen := "unix:///run/infrakit/plugins/"
+	sock := "group.sock"
```
chungers: Is there a reason for splitting this into two separate flags?   The url (with `unix://` as the protocol/scheme) seems to be more common for specifying listen address.   Also, you end up joining the strings to make a url later on in the code.

FrenchBen: The idea behind it is that I can now set 1 listen dir for the entire app while still being able to declare the socket via the `--sock` flag. 
Otherwise the `--listen` flag needs to be over-written every time, since it includes the `group.sock` in its string

FrenchBen:
@wfarner @chungers all tests passed 🎉

wfarner:
First, apologies for the delay in getting back to you on this.  Thanks for the patch, this will be a nice improvement!

Two high-level comments:
1. What is the motivation for having the `_DIR` _and_ the `_LISTEN` vars?  Would it make sense to eliminate the latter for simplicity?
2. There are a number of log statements that were likely useful for development, but should be removed prior to landing on master.

Two other superficial, but necessary, tasks:
1. please [sign your commits](https://github.com/docker/infrakit/blob/master/CONTRIBUTING.md#sign-your-work).  This contribution instruction was added after your PR was sent, so it's no surprise you didn't see it!
2. please update your branch with master.  Our delay in getting back to you has caused some drift.  Sorry about that!

docs/tutorial.md
```diff
@@ -8,53 +8,56 @@ To illustrate the concept of working with Group, Flavor, and Instance plugins, w
 All InfraKit plugins will by default open the unix socket located at /run/infrakit/plugins. Make sure this directory exists on your host:
 
 ```shell
-$ mkdir -p /run/infrakit/plugins/
-$ chmod 777 /run/infrakit/plugins
+# Make sure directory exists for plugins to discover each other (default is: /run/infrakit/plugins/)
+$ mkdir -p /tmp/infrakit/plugins/
```
wfarner: Not terribly important as this doc is only for broad guidance, but `$TMPDIR` is likely more portable than `/tmp`.

FrenchBen: I'll update and rebase 👍

chungers:
I think I'm missing the part about the flag getting overwritten every time.


For each binary there's only one listen flag and it follows the URL
convention.  If you split that then it's strange to have a UNIX protocol in
one flag and a sock file in another.  Also what do we do on a platform that
requires tcp and the flag sock really doesn't make sense...

wfarner:
> For each binary there's only one listen flag and it follows the URL
convention.  If you split that then it's strange to have a UNIX protocol in
one flag and a sock file in another.  Also what do we do on a platform that
requires tcp and the flag sock really doesn't make sense...

The platform issue is a fair criticism.  Could we tackle this by supporting the environment variable only on platforms supporting unix sockets, and using it as a prefix for the `--listen` flag when it is not set?  i.e. i can set `INFRAKIT_PLUGINS_DIR=/tmp/infrakit` and avoid setting `--listen` anywhere else.

wfarner:
FYI i've filed https://github.com/docker/infrakit/issues/214, which would mean we always have a platform-default mechanism for plugin communication and do not need to support configurable modes.

FrenchBen:
Updated PR to have only 1 ENV var `INFRAKIT_PLUGINS_DIR`.
I wanted the flag to be consistent across all cli, but `--dir` was already taken, so `--discovery` was used (open to other suggestions).

Ouput:
```
bash$ build/infrakit-group-default --log 5 &
INFO[0000] Listen: %vunix:///var/folders/kl/4fm2zyxs3_5dd869x992dvvw0000gn/T/infrakit/plugins/group.sock
INFO[0000] Starting discovery
DEBU[0000] Opening: /var/folders/kl/4fm2zyxs3_5dd869x992dvvw0000gn/T/infrakit/plugins
INFO[0000] Starting plugin
INFO[0000] Starting
INFO[0000] Listening on: unix:///var/folders/kl/4fm2zyxs3_5dd869x992dvvw0000gn/T/infrakit/plugins/group.sock
INFO[0000] listener protocol= unix addr= /var/folders/kl/4fm2zyxs3_5dd869x992dvvw0000gn/T/infrakit/plugins/group.sock err= <nil>

bash$ build/infrakit-instance-file --log 5 --dir ./tutorial/ &
INFO[0000] Starting plugin
INFO[0000] Listening on: unix:///var/folders/kl/4fm2zyxs3_5dd869x992dvvw0000gn/T/infrakit/plugins/instance-file.sock
DEBU[0000] file instance plugin. dir= ./tutorial/
INFO[0000] listener protocol= unix addr= /var/folders/kl/4fm2zyxs3_5dd869x992dvvw0000gn/T/infrakit/plugins/instance-file.sock err= <nil>

bash$ build/infrakit-flavor-vanilla --log 5 &
INFO[0000] Starting plugin
INFO[0000] Listening on: unix:///var/folders/kl/4fm2zyxs3_5dd869x992dvvw0000gn/T/infrakit/plugins/flavor-vanilla.sock
INFO[0000] listener protocol= unix addr= /var/folders/kl/4fm2zyxs3_5dd869x992dvvw0000gn/T/infrakit/plugins/flavor-vanilla.sock err= <nil>

bash$ build/infrakit plugin ls
Plugins:
NAME                	LISTEN
flavor-vanilla      	unix:///var/folders/kl/4fm2zyxs3_5dd869x992dvvw0000gn/T/infrakit/plugins/flavor-vanilla.sock
group               	unix:///var/folders/kl/4fm2zyxs3_5dd869x992dvvw0000gn/T/infrakit/plugins/group.sock
instance-file       	unix:///var/folders/kl/4fm2zyxs3_5dd869x992dvvw0000gn/T/infrakit/plugins/instance-file.sock
```

CI output:
```
$ make ci
+ fmt
+ vet
+ lint
+ vendor-sync
+ vendor-check
+ coverage
?   	github.com/docker/infrakit/cmd/cli	[no test files]
?   	github.com/docker/infrakit/cmd/group	[no test files]
ok  	github.com/docker/infrakit/discovery	1.048s	coverage: 78.0% of statements
?   	github.com/docker/infrakit/example/flavor/swarm	[no test files]
?   	github.com/docker/infrakit/example/flavor/vanilla	[no test files]
?   	github.com/docker/infrakit/example/flavor/zookeeper	[no test files]
?   	github.com/docker/infrakit/example/instance/file	[no test files]
?   	github.com/docker/infrakit/example/instance/terraform	[no test files]
?   	github.com/docker/infrakit/example/instance/vagrant	[no test files]
?   	github.com/docker/infrakit/plugin	[no test files]
ok  	github.com/docker/infrakit/plugin/flavor/swarm	1.039s	coverage: 73.2% of statements
?   	github.com/docker/infrakit/plugin/flavor/vanilla	[no test files]
?   	github.com/docker/infrakit/plugin/flavor/zookeeper	[no test files]
ok  	github.com/docker/infrakit/plugin/group	2.142s	coverage: 85.5% of statements
ok  	github.com/docker/infrakit/plugin/group/types	1.027s	coverage: 43.5% of statements
?   	github.com/docker/infrakit/plugin/group/util	[no test files]
?   	github.com/docker/infrakit/plugin/instance/vagrant	[no test files]
ok  	github.com/docker/infrakit/plugin/util	1.038s	coverage: 61.0% of statements
?   	github.com/docker/infrakit/spi/flavor	[no test files]
?   	github.com/docker/infrakit/spi/group	[no test files]
ok  	github.com/docker/infrakit/spi/http/flavor	1.045s	coverage: 91.2% of statements
ok  	github.com/docker/infrakit/spi/http/group	1.097s	coverage: 94.3% of statements
ok  	github.com/docker/infrakit/spi/http/instance	1.083s	coverage: 91.4% of statements
?   	github.com/docker/infrakit/spi/instance	[no test files]
```

cmd/group/main.go
```diff
@@ -127,7 +133,12 @@ func main() {
 		},
 	})
 
-	cmd.Flags().StringVar(&listen, "listen", listen, "listen address (unix or tcp) for the control endpoint")
+	log.Info("Adding cmd binds")
+	cmd.Flags().String("dir", discoveryDir, "Dir path for plugin discovery")
+	// Bind Pflags for cmd passed
+	viper.BindEnv("dir", "INFRAKIT_PLUGINS_DIR")
+	viper.BindPFlag("dir", cmd.Flags().Lookup("dir"))
+	cmd.Flags().String("name", name, "listen socket name for the control endpoint")
```
wfarner: Help text should be something like `"plugin name to advertise"`.  There's a handful of other identical lines that need the same change.

docs/tutorial.md
```diff
@@ -9,20 +9,23 @@ All InfraKit plugins will by default open the unix socket located at /run/infrak
 exists on your host:
 
 ```shell
-$ mkdir -p /run/infrakit/plugins/
-$ chmod 777 /run/infrakit/plugins
+# Make sure directory exists for plugins to discover each other (default is: /run/infrakit/plugins/)
+$ mkdir -p /$TMPDIR/infrakit/plugins/
```
wfarner: No leading `/`.  It's benign, but `$TMPDIR` will include it.

docs/tutorial.md
```diff
@@ -9,20 +9,23 @@ All InfraKit plugins will by default open the unix socket located at /run/infrak
...
 Start the default Group plugin
 
+**_NOTE:_** You can set the listen url as an env variale via `export INFRAKIT_PLUGINS_DIR="/$TMPDIR/infrakit/plugins/"` instead of passing it as a flag. The default socket will be used, unless the `--name` flag is passed, when the env variable is set. 
+
 ```shell
+$ export INFRAKIT_PLUGINS_DIR="/$TMPDIR/infrakit/plugins/"
 $ build/infrakit-group-default --log 5
 INFO[0000] Starting discovery
-DEBU[0000] Opening: /run/infrakit/plugins
+DEBU[0000] Opening: /$TMPDIR/infrakit/plugins
```
wfarner: This is exactly my worry with including console output, rogue find/replace leads to lies!  In reality `$TMPDIR` should be expanded (ditto for numerous other lines).  I'll let you choose how to proceed, i think the console output needs refining anyways, so it's likely to receive another pass soon regardless.

cmd/group/main.go
```diff
@@ -78,7 +84,12 @@ func main() {
 				return flavor_client.PluginClient(callable), nil
 			}
 
-			_, stopped, err := util.StartServer(listen, group_server.PluginServer(
+			log.Infoln("Starting plugin")
```
wfarner: These 3 `log` statements can go away, since removed on master.

PR #195, opened 2016-10-05
A simple AWS EC2 Instance plugin with discovery

chungers:
Here is a minimal EC2 instance plugin
  + It can discovery the subnet, security groups, instance type, image on the instance where
  the plugin is running...  making it easy (nearly no input) to 'clone' more nodes into an existing
  subnet.
  + You can always provide actual values.  The input is just the [`RunInstancesInput`](http://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#RunInstancesInput) in the AWS API.
  + There's a `README.md` on how to use, with example JSONs -- one for the `Properties`
  block, one with a flavor plugin for a full group specification.

This plugin is pretty minimal but the user can get started very quickly, once a host in a subnet in a private VPC is set up.  So it's useful and serves as another plugin example.


wfarner:
Closing this PR, as we'll be putting the AWS plugin in another repository.  Feel free to reopen if you disagree.

chungers:
Agreed.  Going forward plugins should reside in separate git repos and not in the main (except for poc / examples).

Issue #196, opened 2016-10-06
Add DCO check, maintainers file, contributing.md


thaJeztah:
noticed some PRs were merged without a sign-off, and there's no contributing.md or maintainers file present in this repository. Some boilerplating can be found here https://github.com/docker/opensource/tree/master/project-template

wfarner:
Thanks for pointing this out!  We'll hold on accepting PRs until we have this in place.

thaJeztah:
I opened https://github.com/docker/infrakit/issues/209 to add the necessary information; PR's can be merged if they contain a proper DCO sign-off (but having the check automated would probably make it easier)

wfarner:
Looks like we're all set here - the automated check should be configured.  Thanks again for the PR @thaJeztah!

thaJeztah:
Great!

PR #206, opened 2016-10-08
Fix zookeeper example to use valid JSON in README

garethr:
This was fixed in https://github.com/docker/infrakit/pull/197 but only in the accompanying JSON file, not in the README.


wfarner:
Apologies for taking so long to catch up to this, we needed to wait for #196 to be resolved.

Can you update your branch to [sign your commit](https://github.com/docker/infrakit/blob/master/CONTRIBUTING.md#sign-your-work)?  Should only take a moment, then we're all squared away on the legal front :-)

wfarner:
This was superceded by #222 while waiting for the signed commit, but thanks any way for pointing this out!

PR #221, opened 2016-10-13
Adding detail about file path.

hekaldama:
Upon first executing `./infrakit/swarm -h` I saw that
tls{cacert,cert,key} said `string`. Wanted to add a note about file path
to disambiguate between the contents of the file(s) and path to file.

Signed-off-by: Adam Avilla <aavilla@yp.com>


wfarner:
Thanks!

PR #229, opened 2016-10-15
CLI: only accept configs by path name rather than stdin

### LINKED ISSUE #217: `infrakit group watch` hangs when no arguments are provided
This command hangs, which is confusing:
```console
$ build/infrakit group watch
```

I suspect this is an artifact of accepting input via stdin.  Perhaps we should eliminate support for configurations via stdin, or do so only by a non-default use of the command.

wfarner:
Additionally, remove CLI support for interacting with Flavor and Instance plugins.  Rationale being that Flavor and Instance plugins are akin to 'private' APIs that we shouldn't steer folks towards.  Instead we should build tooling around the first-class public APIs, which in this case revolve around Groups.  Once i had removed these features, the `cmd/cli/README.md` didn't seem necessary in addition to `README.md` and `docs/tutorial.md`.

Closes #217

### DIFF
--- a/cmd/cli/README.md
+++ b/cmd/cli/README.md
@@ -21,7 +21,7 @@ instance-file       	~/.infrakit/plugins/instance-file
 ```
 
 Once you know the plugins by name, you can make calls to them.  For example, the instance plugin
-`instance-file` is a pPlugin that "provisions" instances by writing the instructions to
+`instance-file` is a Plugin that "provisions" instances by writing the instructions to
 a file in a local directory.
 
 You can access the following plugins and their methods via command line:
--- a/cmd/cli/README.md
+++ b/cmd/cli/README.md
@@ -1,11 +1,8 @@
 InfraKit CLI
 ============
 
-This is a CLI for working with various infrakit plugins.  In the simplest form, an InfraKit plugin
-is simply a daemon that communicates over unix domain sockets.  InfraKit plugins can find each
-other by the socket files in a common directory.  The CLI uses the common directory as a discovery
-mechanism and offers various subcommands for working with plugins. In general, plugin methods are
-exposed as verbs and configuration JSON can be read from local file or standard input.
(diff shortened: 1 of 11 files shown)


chungers:
I strongly prefer leaving the CLI for instance and, to a lesser degree, the flavor plugins.  This is because the CLI has been very useful for plugin developers.  These allow testing of individual plugins in isolation and has been very useful in my personal development of the terraform plugin and others in the works. 

Furthermore, the CLI is meant to be a developer tool and not end-user tool, because there are plans to have a demux /front-end for managing configs for multiple groups and other infrastructure resource types.  IMHO removing CLI for instance and flavor makes life harder for plugin developers than simplification to improve UX.

chungers:
Sorry fat-fingered and closed the PR by mistake.  Re-opened.  Here's an Issue I created to track the design and implementation for a better UX / frontend that will hopefully solve some of the problems this PR is aiming to fix.  #230

wfarner:
Fair enough.  Let's keep this in mind as we continue to evolve the documentation.  At present the developer-focused nature of the CLI and risk of interacting directly with Instance plugins is not clear.  Hopefully as we break ground with #230 (thanks for filing!) that will become obvious.

I'll update this PR to restore those commands.

chungers:
Thanks.  I agree that docs need to reflect this.

Also another point it occurred to me..  since we are removing tcp support in #228 removing CLI for instance and flavor will make it very difficult to develop since popular tools like curl and postman don't work with unix sockets.  So the CLI are the only means for developers to interactively test the plugins.

wfarner:
Not that it changes the direction we take, but FWIW curl definitely works with unix sockets.

http://superuser.com/questions/834307/can-curl-send-requests-to-sockets

wfarner:
Latest revision restores tools, focuses on file-only config input.

PR #241, opened 2016-10-18
Building InfraKit in containers

### LINKED ISSUE #238: Build binaries in a docker container
Some users have come across issues when building infrakit.  We can improve the build UX by doing the build inside docker containers.

Some things to consider... 
  + Support cross compilation.  Linux is the default target when building but the user's local dev platform is likely not Linux, so
  + Support GOOS=darwin or whatever the user's local development platform is.  This is necessary since we don't require docker to be runtime.

This gets into the topic of whether we should provide Docker containers to simplify experimentation and running of the binaries.  It's outside the scope of this Issue unless people feel otherwise.  Please comment here.

chungers:
This PR addresses #238 to provide an easier option to building InfraKit:

  + A new Makefile target `make build-in-container` is added.
  + Builds the binaries inside a Docker container for deterministic Go compiler version and proper project directory layout.
  + Allows setting of `GOOS` and `GOARCH` for cross-compilation.  By default, it introspects the user's development environment (via `go env GOOS`) to set the environment variables so that even when using a Linux container, the produced binary will run on the user's local machine.
  + Because `govendor sync` can be time-consuming, this build target uses a combination of a data volume container (named with the hash of `vendor/vendor.json`) and `docker run --volumes-from`.  This effectively caches the vendored source -- in named containers -- across builds, so sync is done only once, if the hash of `vendor/vendor.json` changes.

Since this is done as a new build target, there is no impact to current workflow, and those who want to try this only needs to do
```
make build-in-container
```

Finally, a simple Docker container that has the 'reference' environment (with Golang 1.7.1, gomock, govendor, golint pre-installed) is made from `dockerfiles/Dockerfile.devtool-go-1.7.1`.  This image is reused and spares the user of the hassles of setting up environments.

Closes #238


Makefile
```diff
@@ -24,6 +24,31 @@ AUTHORS: .mailmap .git/HEAD
...
+# volume container across multiple builds, as long as the vendoring dependencies
+# don't chnage.  When vendor.json is updated, a new hash results in a new container
+# which is reused across builds again.  If you bulid this target, vendor/ directory
+# will not be changed.  Instead the vendored files will be in the container's fs namespace.
+build-in-container:
+	@echo "+ $@"
+	docker build -t infrakit-build:${VENDOR_HASH} -f ${CURDIR}/dockerfiles/Dockerfile.build .
+	-docker run --name infrakit-build-${VENDOR_HASH} \
+		-e GOOS=${GOOS} -e GOARCCH=${GOARCH} \
+		infrakit-build:${VENDOR_HASH}
```
thaJeztah: How are we cleaning up the anonymous volume afterwards?

justincormack: @thaJeztah when your computer fills up infrakit will provision another one!

justincormack:
Now the deps are vendored, this is much simpler...

Makefile
```diff
@@ -24,6 +24,31 @@ AUTHORS: .mailmap .git/HEAD
...
+
+# First we create a container that is versioned (via the vendor.json hash) that has
+# all the downloaded vendor dependencies.  Then we use that container as the data
+# volume container across multiple builds, as long as the vendoring dependencies
+# don't chnage.  When vendor.json is updated, a new hash results in a new container
+# which is reused across builds again.  If you bulid this target, vendor/ directory
+# will not be changed.  Instead the vendored files will be in the container's fs namespace.
+build-in-container:
+	@echo "+ $@"
+	docker build -t infrakit-build:${VENDOR_HASH} -f ${CURDIR}/dockerfiles/Dockerfile.build .
```
psftw: subsequent builds cause container name conflict
```shell
docker: Error response from daemon: Conflict. The name "/infrakit-build-86c1e7cb1ae5c6f78213c2bd113098837bcd8a9f" is already in use by container fd9f5dee8c2dca03fe5a33cb08381624dde47e0a197a7f5d27153f4adbac0208. You have to remove (or rename) that container to be able to reuse that name..
See 'docker run --help'.
Makefile:41: recipe for target 'build-in-container' failed
make: [build-in-container] Error 125 (ignored)
```

Makefile
psftw: An alternative implementation that's a little quicker and doesn't require volumes:
```diff
diff --git a/.dockerignore b/.dockerignore
new file mode 100644
index 0000000..d097005
--- /dev/null
+++ b/.dockerignore
@@ -0,0 +1,4 @@
+build
+dockerfiles
+vendor
+!vendor/vendor.json
diff --git a/Makefile b/Makefile
index da5ebd0..72dc343 100644
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,14 @@ build-in-container:
                --volumes-from infrakit-build-${VENDOR_HASH} \
                infrakit-build:${VENDOR_HASH} make build-binaries
 
+build-in-container-simple:
+       @echo "+ $@"
+       docker build -t infrakit-build -f ${CURDIR}/dockerfiles/Dockerfile.build-simple .
+       docker run --rm \
+               -e GOOS=${GOOS} -e GOARCCH=${GOARCH} \
+               -v ${CURDIR}/build:/go/src/github.com/docker/infrakit/build \
+               infrakit-build make build-binaries
+
 vet:
        @echo "+ $@"
        @go vet $(PKGS)
diff --git a/dockerfiles/Dockerfile.build-simple b/dockerfiles/Dockerfile.build-simple
new file mode 100644
index 0000000..a725800
--- /dev/null
+++ b/dockerfiles/Dockerfile.build-simple
@@ -0,0 +1,9 @@
+FROM infrakit/devtool:go-1.7.1
+
+WORKDIR /go/src/github.com/docker/infrakit
+
+COPY Makefile /go/src/github.com/docker/infrakit/Makefile
+COPY vendor/vendor.json /go/src/github.com/docker/infrakit/vendor/vendor.json
+RUN make vendor-sync
+
+COPY . /go/src/github.com/docker/infrakit
```

psftw: ^ oops, still getting used to GitHub reviews, meant to post this in a separate "conversation"

chungers:
Now that sources are vendored thanks to #243 I will make another update.   We won't need the data volume container anymore...

dockerfiles/Dockerfile.build
```diff
@@ -0,0 +1,22 @@
+FROM golang:1.7.1-alpine
+
+# A container for building InfraKit
+
+RUN apk add --update git tree make
+
+ENV GOPATH=/go
+ENV PATH=/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+
```
justincormack: I think you need `ENV CGO_ENABLED=0` to build static binaries too (else they are linked against Musl libc)

wfarner:
LGTM

dockerfiles/Dockerfile.build
chungers: Thanks!  Adding that.

PR #242, opened 2016-10-18
Add a Flavor plugin that composes other flavor plugins

wfarner:
I also noticed that `--name` was not broken in several plugins, fixed that as well.


example/flavor/combo/README.md
```diff
@@ -0,0 +1,91 @@
+InfraKit Flavor Plugin - Combo
+==============================
+
+A [reference](../../../README.md#reference-implementations) implementation of a Flavor Plugin that supports composition
+of other Flavors.
+
+The Combo plugin allows you to use Flavors as mixins, combining their Instance properties.
+
```
wfarner: Sure.

plugin/flavor/zookeeper/flavor.go
```diff
@@ -167,7 +167,7 @@ func (s zkFlavor) Prepare(
 			return spec, errors.New("Manager nodes require an assigned logical ID")
 		}
 
-		spec.Init = generateInitScript(properties.UseDocker, allocation.LogicalIDs, *spec.LogicalID)
+		spec.Init += "\n" + generateInitScript(properties.UseDocker, allocation.LogicalIDs, *spec.LogicalID)
```
chungers: Why a newline here -- does this depend on how this flavor is used (by itself or in a combo)?

wfarner: Thanks for catching that, the newline should be conditional.  I meant to check if the init script introduced a shebang and it does, so the newline should only be included when appending.

example/flavor/combo/flavor.go
```diff
@@ -0,0 +1,63 @@
...
+		if err != nil {
+			return instance, err
+		}
+
+		var props json.RawMessage
+		if pluginSpec.Properties != nil {
+			props = *pluginSpec.Properties
+		}
+
+		instance, err = plugin.Prepare(props, instance, allocation)
```
chungers: So this looks like a plugin can overwrite the output of previous plugins?  I am not sure what else can be done but it may be useful to call that out and we should think about some recommended behavior for plugins in their `Prepare` method.

wfarner: An alternative - i could give each plugin a copy of the `instance.Spec` and define merging behavior once and for all in the Combo plugin.  This has its own caveats, but at least behavior is predictable across different Flavors.

chungers: I think that would be nice.   I can see how some plugin could just overwrite things like Init, which really should be properly merged (like removing `#!`), etc.   Tags could also be merged in a predictable way.

chungers: Also for logical ID and attachment ids....  yeah definitely we can have issues with stuff getting clobbered and maybe in those cases the combo should complain loudly or have some behavior controllable in the config itself (maybe later).

PR #243, opened 2016-10-18
Switch to trash for vendor management

### LINKED ISSUE #189: Consider govendor alternatives
SwarmKit just switched to `vndr` for vendoring (https://github.com/docker/swarmkit/pull/1546). Let's consider using that.

Don't switch right away: it would be good to get feedback.

@LK4D4

wfarner:
Switches the project to [trash](https://github.com/rancher/trash) to manage dependencies.  As a result, vendored sources are now checked in.  I'm now considering checked-in vendored sources a necessary evil to eliminate vendor tooling from revealing itself to the casual `infrakit` user/contributor, and speeding up builds in fresh environments.  Only contributors looking to change dependencies will need to install `trash`.

`trash` aims to be much less sophisticated than `govendor`, which i'm hoping will eliminate some of the pain we've experienced when modifying dependencies.  See notes in `vendor.conf` for how to use `trash`.

As expected, this patch is huge.  Luckily, the most relevant changes are at the top, with the vast majority in `vendor/`.

Closes #189


chungers:
There are 738 files and I only wanted to see the Makefile changes...  github is giving me problems.  any chance they can be split up so that one commit has all the vendored sources?

wfarner:
@chungers done

chungers:
The fact that [trash](https://github.com/rancher/trash) can distinguish import package name from git repo is nice.   How is it different from https://github.com/LK4D4/vndr ?

wfarner:
> How is it different from https://github.com/LK4D4/vndr ?

In broad strokes: more project activity overall: releases, contributors, issues.  Worked on the first try, so i didn't go any further on a comparative analysis.

chungers:
I think it's an improvement over govendor.  That said, I am not super crazy about checking in all the vendored sources since there's no mechanism that guarantees the checked in source are indeed from the repo/shas listed in the manifest (`vendor.conf`).  The best way to guarantee integrity of the vendored source is to fetch them from repos at build time, even though the initial UX isn't optimal.

I wonder if it make sense to address the problems separately...  1. have a simpler tool than govendor and 2. use containers/ data volumes to cache dependencies across builds (like #241) so we can do without checking in all the vendored sources and still have reasonable build time / UX.

Can people chime in on this?

wfarner:
> I am not super crazy about checking in all the vendored sources since there's no mechanism that guarantees the checked in source are indeed from the repo/shas listed in the manifest

Checking in vendored sources doesn't eliminate the ability to validate them.  We can add a vendor check to `make ci` without too much hassle, though i haven't seen any strong indication that this is common (disclaimer - i haven't searched too much).

chungers:
@justincormack @thaJeztah @stevvooe 

related #189 #241

chungers:
Yeah I am ok with checking in source now that I see swarmkit also checks in the source in addition to using vndr.  

We can add a check later, or at least be vigilant when some big PR lands with a bunch of vendor sources and stuff all rolled into one.

thaJeztah:
fwiw, docker/docker has a validate-vendor to check if vendored files actually match the ones specified in the vendor-file, this prevents accidental changes in them as part of a PR, or incorrectly vendored files; https://github.com/docker/docker/blob/master/hack/make/validate-vendor

thaJeztah:
Looks like there's a merge commit in this PR ("Merge branch 'master' into vendor-trash"), so may need a rebase

thaJeztah:
SGTM

wfarner:
@thaJeztah 
> Looks like there's a merge commit in this PR ("Merge branch 'master' into vendor-trash"), so may need a rebase

We squash our PR commits, so no worries there.

wfarner:
I'm going to go ahead and merge this as we have quorum.  We can always revert or course-correct if we decide this isn't the approach we want.

PR #246, opened 2016-10-20
Add a script to test the tutorial command sequence

wfarner:
I've found myself manually doing this to avoid breaking the tutorial when making other changes.  This allows me to be lazy and validate automatically :-)

If this is something we want to formalize, it would be worth including in the `make ci` target.

### DIFF
--- a/plugin/group/quorum.go
+++ b/plugin/group/quorum.go
@@ -50,6 +50,7 @@ func (q *quorum) Stop() {
 func (q *quorum) Run() {
 	ticker := time.NewTicker(q.pollInterval)
 
+	q.converge()
 	for {
 		select {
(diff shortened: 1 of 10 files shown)


plugin/group/quorum.go
```diff
@@ -50,6 +50,7 @@ func (q *quorum) Stop() {
 func (q *quorum) Run() {
 	ticker := time.NewTicker(q.pollInterval)
 
+	q.converge()
```
wfarner: This is scratching an itch, the result is instances are created immediately when first `watch`ing a group.  For the script, it means a shorter `sleep`.

chungers:
Nice work.  Definitely should do this more.

PR #247, opened 2016-10-21
Use flavor.Health as feedback during rolling updates

### LINKED ISSUE #234: Group updater: incorporate flavor.Healthy as a signal for safer updates
Addresses this TODO in `rollingupdate.go`:
> // TODO(wfarner): Poll ProvisionHelper.Healthy here.

wfarner:
I believe i have taken reasonable and conservative measures to err on the side of safety in the course of an update.  There are many ways to make updating very sophisticated (and complicated), but for a first cut i have erred on the side of safety and simple/predictable behavior.

_Echoing comments on behavior from the source_
Rolling update implementation - ignore health of 'old' instances:
```go
// The update is only concerned with instances being created in the course of the update.
// The health of instances in any other state is irrelevant.  This behavior is important
// especially if the previous state of the group is unhealthy and the update is attempting to
// restore health.
```

Rolling update implementation - handling `Healthy()` results:
```go
// Now that we have isolated the instances with the new configuration, check if they are all
// healthy.  We do not consider an update successful until the target number of instances are
// confirmed healthy.
// The following design choices are currently implemented:
//
//   - the update will continue indefinitely if one or more instances are in the
//     flavor.UnknownHealth state.  Operators must stop the update and diagnose the cause.
//
//   - the update is stopped immediately if any instance enters the flavor.Unhealthy state.
//
//   - the update will proceed with other instances immediately when the currently-expected
//     number of instances are observed in the flavor.Healthy state.
```

Flavor combo `Healthy()` implementation:
```go
// The overall health of the flavor combination is taken as the 'lowest common demoninator' of the configured
// flavors.  Only flavor.Healthy is reported if all flavors report flavor.Healthy.  flavor.Unhealthy or
// flavor.UnknownHealth is returned as soon as any Flavor reports that value.
```

Closes #234

### DIFF
--- a/plugin/group/rollingupdate.go
+++ b/plugin/group/rollingupdate.go
@@ -87,6 +87,9 @@ func (r *rollingupdate) waitUntilQuiesced(pollInterval time.Duration, expectedNe
 			//
 			numHealthy := 0
 			for _, inst := range matching {
+				// TODO(wfarner): More careful thought is needed with respect to blocking and timeouts
+				// here.  This might mean formalizing timeout behavior for different types of RPCs in
+				// the group, and/or documenting the expectations for plugin implementations.
 				switch r.scaled.Health(inst) {
 				case flavor.Healthy:
 					numHealthy++
--- a/plugin/group/rollingupdate.go
+++ b/plugin/group/rollingupdate.go
@@ -2,7 +2,9 @@ package group
 
 import (
 	"errors"
+	"fmt"
 	log "github.com/Sirupsen/logrus"
+	"github.com/docker/infrakit/spi/flavor"
 	"github.com/docker/infrakit/spi/instance"
 	"sort"
 	"time"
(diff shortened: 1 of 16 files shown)


spi/flavor/spi.go
```diff
@@ -6,6 +6,20 @@ import (
 	"github.com/docker/infrakit/spi/instance"
 )
 
+// Health is an indication of whether the Flavor is functioning properly.
+type Health int
+
+const (
+	// UnknownHealth indicates that the Health cannot currently be confirmed.
+	UnknownHealth Health = iota
```
chungers: Nit -- `Unknown` better?   The type is already `Health`.

plugin/group/rollingupdate.go
```diff
@@ -64,20 +66,41 @@ func (r *rollingupdate) waitUntilQuiesced(pollInterval time.Duration, expectedNe
...
+			//     flavor.UnknownHealth state.  Operators must stop the update and diagnose the cause.
+			//
+			//   - the update is stopped immediately if any instance enters the flavor.Unhealthy state.
+			//
+			//   - the update will proceed with other instances immediately when the currently-expected
+			//     number of instances are observed in the flavor.Healthy state.
+			//
+			numHealthy := 0
+			for _, inst := range matching {
+				switch r.scaled.Health(inst) {
```
chungers: This call to get health reaches out to the flavor plugin to get health, which is a RPC call.  How do we handle the cases when this call blocks for a long time?   It would seem that the call would block for a long time especially when often 'unhealthy' means timing out after a long wait.  Also this is iterating over several instances, one by one, I wonder if this will cause problems within the polling loop. 

Maybe N go routines with timeout / cancellation with timeout < pollInterval?

spi/flavor/spi.go
wfarner: Sounds good.

plugin/group/rollingupdate.go
wfarner: This is an issue for all inter-plugin calls, and something i think we should take a general stance on rather than addressing in an ad hoc manner.

Given the way we are treating plugins (effectively as trusted DLLs), i think our stance should be that network overhead is negligible and plugins conform to (yet-to-be-published) guidelines on blocking when handling their interface methods.

wfarner: Left a TODO for us to reassess.

Issue #244, opened 2016-10-19
Introduce a plugin type to manage Group state


wfarner:
The default Group plugin currently only has in-memory state for the Groups being watched.  Introduce a new plugin type that the Group plugin is configured with, which delegates state management.

chungers:
What's the SPI for this?   Is this something that needs to be applied more broadly to other plugin types as well?   I was thinking an 'environment' plugin along these lines -- one that encapsulates state, plugin discovery and leader / non-leader into one common plugin that is run as a singleton per installation.

wfarner:
> What's the SPI for this?

That's what i intended to sort out in the issue discussion :-)

> Is this something that needs to be applied more broadly to other plugin types as well?

You mean state for other plugins?  I was planning to start small and focus on flexible primitives, but focus on wiring it only to the group plugin as that's the immediate need.

> I was thinking an 'environment' plugin along these lines -- one that encapsulates state, plugin discovery and leader / non-leader into one common plugin that is run as a singleton per installation.

Those seem like distinct behaviors so my inclination was to introduce them as separate plugins.

chungers:
Ok - I think it's reasonable to keep it simple and just focus on helping Group manage state.

Since the group plugin doesn't have a strict schema, I am thinking along the lines of simple POST/PUT/GET/DELETE by path (namespace) with the payload as a blob with a content-type (just to make it general).  This is pretty generic and can be easily backed by a variety of existing tooling - from local disk (the default) to kv stores or full-on document databases (eg. couchdb).  

(Path, value) are nice since we can define namespaces for the plugins when there are many groups with different implementations.

However I think it'd be nice for the group to store its state in one atomic document / blob.  So while the state store supports (path, value), a given plugin instance in practice always CRUD with a fix key/path.  So maybe something (the "state" plugin) should assign the key and the plugin only needs to hand over a blob to persist or do a `get` to read.

On the Go side, I'd imagine something that does upsert / delete?   Deletion can be a funny thing and should warrant discussion.

Or do you have something more narrowly defined?

wfarner:
> Or do you have something more narrowly defined?

I have nothing defined.  I tend to file issues before shaping the design, so what you see is what you get.

chungers:
I am going to introduce a PR that has only a simple Golang API to save / load a blob of some sort.  This follows from the comment earlier  https://github.com/docker/infrakit/issues/244#issuecomment-254981859

[...]

PR #249, opened 2016-10-25
Add an example using the swarm flavor plugin

### LINKED ISSUE #204: ./example/flavor/swarm needs documenting
A README in the style of the tutorial would be very nice.

Extra points for tying it into the Terraform example, showing how to deploy a swarm onto some terraformed instances.

wfarner:
Additionally, this removes shell code to install Docker.  Since we now have the
ability to compose plugins (using the combo flavor plugin), this is best done with composition.

Closes #204

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,8 @@ It knows only about individual instances and nothing about Groups.  Instance is 
 need not be a physical machine at all.
 
 For compute, for example, instances can be VM instances of identical spec. Instances
-support the notions of attachment to auxiliary resources.  Instances are taggable and tags are assumed to be persistent
-which allows the state of the cluster to be inferred and computed.
+support the notions of attachment to auxiliary resources.  Instances may be tagged, and tags are assumed to be
+persistent which allows the state of the cluster to be inferred and computed.
 
 In some cases, instances can be identical, while in other cases the members of a group require stronger identities and
 persistent, stable state. These properties are captured via the _flavors_ of the instances.
@@ -86,14 +86,14 @@ Over time, we would prefer to phase out reference Plugins that appear to provide
 are developed independently.  For this reason, please [file an issue](https://github.com/docker/infrakit/issues/new)
 to start a discussion before contributing to these plugins with non-trivial code.
 
-| plugin                                            | type     | description                             |
-|:--------------------------------------------------|:---------|:----------------------------------------|
-| [swarm](plugin/flavor/swarm)                      | flavor   | runs Docker in Swarm mode               |
-| [vanilla](plugin/flavor/vanilla)                  | flavor   | manual specification of instance fields |
-| [zookeeper](plugin/flavor/zookeeper)              | flavor   | run an Apache ZooKeeper ensemble        |
-| [infrakit/file](./example/instance/file)          | instance | useful for development and testing      |
-| [infrakit/terraform](./example/instance/terraform)| instance | creates instances using Terraform       |
-| [infrakit/vagrant](./example/instance/vagrant)    | instance | creates Vagrant VMs                     |
+| plugin                                             | type     | description                             |
+|:---------------------------------------------------|:---------|:----------------------------------------|
+| [swarm](example/flavor/swarm)                      | flavor   | runs Docker in Swarm mode               |
+| [vanilla](example/flavor/vanilla)                  | flavor   | manual specification of instance fields |
+| [zookeeper](example/flavor/zookeeper)              | flavor   | run an Apache ZooKeeper ensemble        |
+| [infrakit/file](example/instance/file)             | instance | useful for development and testing      |
+| [infrakit/terraform](example/instance/terraform)   | instance | creates instances using Terraform       |
+| [infrakit/vagrant](example/instance/vagrant)       | instance | creates Vagrant VMs                     |
 
 
 #### Supported implementations
@@ -102,7 +102,7 @@ project, so please double-check where the code lives before filing InfraKit issu
 
 | plugin                                                        | type     | description                                           |
 |:--------------------------------------------------------------|:---------|:------------------------------------------------------|
-| [infrakit/group](./cmd/group)                                 | group    | supports Instance and Flavor plugins, rolling updates |
+| [infrakit/group](cmd/group)                                   | group    | supports Instance and Flavor plugins, rolling updates |
 | [docker/infrakit.aws](https://github.com/docker/infrakit.aws) | instance | creates Amazon EC2 instances                          |
 
 Have a Plugin you'd like to share?  Submit a Pull Request to add yourself to the list!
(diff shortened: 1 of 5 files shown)


linsun:
Thank you @wfarner!  I will try it tmr.   Qs:
- I don't quite understand why docker install step is removed (e.g. wget -qO- https://get.docker.com/ | sh) in the  flavor.go.  How would we install docker engine on the host?  Is it because the box file already has docker installed?
- I don't see "type": "manager" specified in the swarm-managers group.json.   I noticed the flavor-swarm isn't used here either.  is that designed?
- how are you proposing to get the IP before the swarm manager group is deployed?
- I think I'm missing how this works, could you explain why swarm node is getting the manager's Ip from swarm plugin's flavor.go?

wfarner:
@linsun it's not exactly straightforward, so i'd be more worried if you didn't have questions :-)

> How would we install docker engine on the host? Is it because the box file already has docker installed?

That's the preference.  Alternatively, you can also install it with shell code using the `vanilla` flavor _before_ the `swarm` flavor, leveraging the flavor [`combo`](https://github.com/docker/infrakit/tree/master/example/flavor/combo) to compose the two

> I don't see "type": "manager" specified in the swarm-managers group.json. I noticed the flavor-swarm isn't used here either. is that designed?

This is a result of me trying to put together a very simple example with minimal code changes.  The Swarm flavor faces a 'bootstrapping' challenge, where exactly one node must run `swarm init`, and must do so _before_ any other manager or worker nodes are created.  InfraKit doesn't have any sort of sequencing like that right now, so i've pushed the initialization and sequencing to the user (create manager group, wait, create worker group).

> how are you proposing to get the IP before the swarm manager group is deployed?

In this case, the IP address is dictated in the `LogicalID` and must be accommodated by the instance plugin.  In the case of Swarm in particular, using stable IP addresses is beneficial as Swarm uses this to identify nodes in the raft cluster.  As a result, instances being updated or re-provisioned causes less raft churn and induces less overall risk in managing the raft state.

> I think I'm missing how this works, could you explain how swarm node is getting the manager's Ip from swarm plugin's flavor.go?

There's two relevant details:
- the swarm plugin is configured to use the manager's IP address for Docker API calls:
```shell
$ build/infrakit-flavor-swarm --host tcp://192.168.2.200:4243
```

- when asked to provision a node, we inject a boot script that uses `ManagerStatus.Addr` from the Docker API to construct the join command:

```
docker swarm join {{.MY_IP}} --token {{.JOIN_TOKEN}}
```

**Stepping back**

This is not the ideal InfraKit/Swarm integration we're picturing.  The ideal setup looks something more like:
- InfraKit runs _on_ the swarm manager nodes, and piggy-backs on swarm's leader election for mutual exclusion
- A bootstrap routine creates the first manager node, runs `swarm init`, and starts InfraKit
- Persistent volumes are used to detach raft state from the compute nodes

There's more work to be done to enable that setup, including #244 and likely some yet-to-be-created issues.

linsun:
@wfarner thanks for answering my questions in detail, much appreciated!  I'm working with VM from IBM Bluemix/Softlayer, where I cannot really assign an IP when I provision a VM today.  One workaround I can think of is to use the vanilla flavor and grab the public ip from user data script (I need to double check to see if the public ip is assigned when the script is executed), and use that to configure the swarm mgr.   If i do that, I'm having trouble to think how to pass the mgr IP to the node as the mgr IP is only configured in the mgr's group.json, not the node's group.json from your example.  In the flavor.go prepare method, you had:

	self, _, err := s.client.NodeInspectWithRaw(context.Background(), nodeInfo.Swarm.NodeID)

which seems to be able to get the nodeInfo that contains the ManagerStatus.Addr, however, how is the mgr's address obtained before the swarm node joins the mgr (which happens a few lines later when generateInitScript() is called)?  This was where I was quite confused....

The combo plugin is an interesting concept, thanks for sharing that!

I agree that if infrakit runs on the mgr node, it will be much easier to discover the mgr, find the join token, watch the node, etc.

wfarner:
@linsun the case of provisioning managers with dynamic IPs is going to be tricky and require some careful thought.  Can i trouble you to file an issue so we can track and continue the discussion?

linsun:
@wfarner good suggestion - opened https://github.com/docker/infrakit/issues/252 so we can continue discussion there!

linsun:
Q @wfarner, if I stand up my swarm mgr (not known the ip till it is up running), and then when I launch the swarm flavor plugin and pass in the mgr ip, similar like ( build/infrakit-flavor-swarm --host tcp://{mgr_ip}:4243), I think I still shall be able to get the node joined the swarm cluster, no?  I plan to give this try with a combo plugin where it uses vanilla flavor to install docker, then use the swarm flavor to join the cluster.

wfarner:
@linsun yup, that should work just fine!

linsun:
@wfarner great, thanks for confirming that!

wfarner:
@linsun another approach worth exploring is running InfraKit directly on the manager node.  This allows you to avoid specifying the IP address.  I'm hoping we will have a demonstration of this kind of behavior in the next week or two.

linsun:
@wfarner Thanks!  I thought about that too... I 'd like to try that after I get the simple path discussed few hours ago working.  If we run infrakit directly on mgr node, perhaps we could develop a simple infrakit flavor to have infrakit installed?

wfarner:
@linsun 

> If we run infrakit directly on mgr node, perhaps we could develop a simple infrakit flavor to have infrakit installed?

My plan was just to use shell code in the vanilla plugin to do this.  As intrigued as i am by the meta nature of an infrakit flavor plugin, i'm struggling to find a benefit over shell code :-)

linsun:
@wfarner Ah, i see.  Shell code is great.   Having a flavor will allow code reuse, as you'll need to install go lang, infrakit including various plugins plus terraform assuming terraform is used for instances.

linsun:
BTW, @wfarner I tried to stand up a swarm node by using build/infrakit-flavor-swarm --host tcp://{mgr_ip}:4243.   It took me quite a while to figure out where it failed, the swarm join cmd failed when I ran manually:   Error response from daemon: Timeout was reached before node was joined. The attempt to join the swarm will continue in the background. Use the "docker info" command to see the current swarm status of your node.

I then discovered port 2377 works fine if I replace the same docker swarm join cmd with port 2377 instead of 4243.  Going to try " build/infrakit-flavor-swarm --host tcp://{mgr_ip}:2377" to see if it works.

linsun:
"build/infrakit-flavor-swarm --host tcp://{mgr_ip}:2377"  failed quickly with "ERRO[0098] Prepare phase failed to fetch Swarm join tokens error=Get http://{mgr_ip}:2377/v1.24/swarm: malformed HTTP response "\x15\x03\x01\x00\x02\x02"...

linsun:
@wfarner -- I've gotten this working on my 2nd try... not sure why the first try didn't work, but I've added a bunch more info/debug statement to help out on 2nd try.  It could be that I didn't wait long enough for the first try as it takes time to get the docker installed plus join swarm cluster.   The ports were fine after all!  Going to try run infrakit on mgr node tmr.

wfarner:
@linsun that's great to hear!  I've also made quite a bit of progress on my bootstrapping PR.  Still a work in progress, but you can see it over here: https://github.com/docker/infrakit.aws/pull/17

PR #251, opened 2016-10-27
Add support for external template to vagrant

FrenchBen:
Updated the Vagrant plugin to conditionally support using a vagrant file processed as a go template.
This allows for greater flexibility in the declaration of vagrant instance specifics, by passing the `--template` to the vagrant plugin

cc @wfarner @chungers

### DIFF
--- a/example/instance/vagrant/main.go
+++ b/example/instance/vagrant/main.go
@@ -1,12 +1,13 @@
 package main
 
 import (
+	"os"
+
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/infrakit/cli"
 	"github.com/docker/infrakit/plugin/instance/vagrant"
 	instance_plugin "github.com/docker/infrakit/spi/http/instance"
 	"github.com/spf13/cobra"
-	"os"
 )
 
 func main() {
@@ -14,6 +15,7 @@ func main() {
 	var name string
 	var logLevel int
 	var dir string
+	var template string
 
 	cmd := &cobra.Command{
 		Use:   os.Args[0],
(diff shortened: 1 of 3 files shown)


plugin/instance/vagrant/instance.go
```diff
@@ -74,6 +73,10 @@ func (v vagrantPlugin) Provision(spec instance.Spec) (*instance.ID, error) {
 	}
 
 	templ := template.Must(template.New("").Parse(vagrantFile))
+	if _, err := os.Stat(v.VagrantfilesDir + "vagrant.tmpl"); err == nil {
```
wfarner: I'm not a big fan of coupling the template with the data dir.  For example, it's nice if the dir could be considered disposable (but not the template).  How about a `--template` argument to the plugin instead?

FrenchBen: I'll go ahead and implement that

example/instance/vagrant/main.go
```diff
@@ -1,27 +1,29 @@
...
+	var template string
 
 	cmd := &cobra.Command{
 		Use:   os.Args[0],
 		Short: "Vagrant instance plugin",
 		Run: func(c *cobra.Command, args []string) {
 
 			cli.SetLogLevel(logLevel)
-			cli.RunPlugin(name, instance_plugin.PluginServer(vagrant.NewVagrantPlugin(dir)))
+			cli.RunPlugin(name, instance_plugin.PluginServer(vagrant.NewVagrantPlugin(dir, template)))
```
wfarner: Let's go ahead and read the template file here, and inject the template contents (or the `const` if the template arg was not specified).  This way we fail fast if the file doesn't exist, and avoid the branch in `instance.go`.

The next step is to accept an abstract JSON schema for the instance plugin properties (so that when i write a template i can define my own variables).  But that can be an exercise for another day.

FrenchBen: You mean pass the `* Template` to the plugin?

wfarner: I was thinking `string`, but `*Template` is even better since you can then fail fast on a bogus template. :+1:

wfarner:
Build timed out, this is a known issue with a test hanging.  I've kicked off a rebuild, looking at the test issue separately.

wfarner:
Thanks for the patch!

PR #253, opened 2016-10-27
Use json-rpc for communication between plugins

### LINKED ISSUE #245: Use json-RPC for inter-plugin communication
Our scaffolding for communication already resembles the APIs to `net/rpc`, which can be adapted to JSON-RPC with `net/rpc/jsonrpc`.  The switch would allow us to offload some custom code and move to a standard wire format as well.

Also weigh `net/rpc/jsonrpc` against other libraries like `github.com/gorilla/rpc/json`.

chungers:
New server / client implementation using the standard JSON-RPC package.

This simplifies the client / server development of plugins and allows interoperability with other systems that support JSON-RPC.

closes #245


cmd/cli/plugin.go
```diff
@@ -24,10 +24,10 @@ func pluginCommand(plugins func() discovery.Plugins) *cobra.Command {
 			}
 
 			if !quiet {
-				fmt.Printf("%-20s\t%-s\n", "NAME", "LISTEN")
+				fmt.Printf("%-20s\t%-8s\t%-s\n", "NAME", "PROTOCOL", "LISTEN")
 			}
 			for k, v := range entries {
-				fmt.Printf("%-20s\t%-s\n", k, v.String())
+				fmt.Printf("%-20s\t%-8s\t%-s\n", k, v.Protocol, v.Address)
```
wfarner: Seems like name and address (path) are the only details relevant to users for now...so we should nix `v.Protocol`.

We can (and should) surface this when we support it in a usable form.

rpc/server.go
```diff
@@ -60,6 +66,47 @@ func StartPluginAtPath(socketPath string, endpoint http.Handler) (chan<- struct{
...
+		*userInitiated = true
+		listener.Close()
+	}()
+
+	go func() {
+	serverLoop:
+		for {
+			conn, err := listener.Accept()
+			if err != nil {
+				break serverLoop
```
wfarner: This may be personal taste, but i'd rather see an unqualified `break` here since the loop label isn't needed, or a loop control variable:

```go
listening := true
for listening {
  ...
  if err != nil {
    listening = false
    continue
  }
}
```

Though of these two, i lean towards the unqualified `break`.

cmd/cli/plugin.go
chungers: Sounds good.  Removing for now.

rpc/server.go
chungers: I think it's clear enough in this case even without the label (no `select` nested in the `for` loop).  Ok. Removing.

chungers:
I will add the tests back...

chungers:
also ran `scripts/tutorial_test` and everything passed.

PR #256, opened 2016-10-28
Remove flaky TestStopUpdate

wfarner:
This test suffers from starvation, leading to intermittent
timeouts of the test.  The culprit is the poll loop in the
updater, which can be stopped.  Because the test uses a very
short timer (for test speed), the stop channel can become
starved in the polling loop


wfarner:
Closing this.  Turns out this does not appear to be starvation after all.  Fix incoming.

PR #257, opened 2016-10-28
Vagrant Abstract JSON

FrenchBen:
#Followup to PR #251 allows JSON to be defined as the user desires, with the only requirements being:
* Box
* Memory (added if missing)
* CPU (added if missing)

cc @wfarner @chungers


wfarner:
Thanks!

PR #258, opened 2016-10-28
Fix deadlocks in rolling update tests

wfarner:
Fixes a few independent issues:
- several test cases weren't unwatching groups, leaving extra goroutines running (benign, but untidy)
- not preventing multiple writes to unbuffered `healthChecksStarted`, causing a deadlock
- holding a lock in `scaledGroup` while calling plugin functions, opening possibility of a deadlock


plugin/group/scaled.go
```diff
@@ -38,34 +38,42 @@ func (s *scaledGroup) changeSettings(settings groupSettings) {
 	s.settings = settings
 }
 
-func (s *scaledGroup) CreateOne(logicalID *instance.LogicalID) {
+// latestSettings gives a point-in-time view of the settings for this group.  This allows other functions to
+// safely use settings and make calls to plugins without holding the lock.
+func (s *scaledGroup) latestSettings() groupSettings {
```
chungers: I am not sure if this does much.  The lock is held as soon as the settings, which you want only for read, is copied and returned on the stack.   Ideally, you want to block reads only when an update of the settings is being updated.  So I think the `scaledGroup.lock` should be a `RWMutex`.  Then this function isn't needed and you just call `lock.RLock() / defer lock.RUnlock()`.   Concurrent reads won't block if no update on the setting is in progress (as signaled by `lock.Lock() / lock.Unlock()`).

wfarner: What you're describing is a different policy, and would have the result of blocking writes (updates) while the read lock is held.  As it stands, the patch decouples locking of the settings from the code acting on the settings.  My intention was to not allow long-running RPCs to block other operations.

chungers: Ok.  We'd have to address long-running RPCs separately.

PR #260, opened 2016-10-28
group: add DescribeGroups()

AkihiroSuda:
This PR adds `DescribeGroups()`, which list up the available groups.

Added CLI as well.

Usage:
```console
$ infrakit group describe-groups
ID
aws-example
```
Signed-off-by: Akihiro Suda <suda.akihiro@lab.ntt.co.jp>

### DIFF
--- a/cmd/cli/group.go
+++ b/cmd/cli/group.go
@@ -232,8 +232,29 @@ func groupPluginCommand(plugins func() discovery.Plugins) *cobra.Command {
 			return err
 		},
 	}
+	describeGroups := &cobra.Command{
+		Use:   "describe-groups",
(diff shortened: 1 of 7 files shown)


wfarner:
Looks great, thanks for the patch!  Just a heads up - i'll probably follow up to rename some things (we should really call this `describe` and rename `describe` to `describe-update`).

AkihiroSuda:
Yes, I was thinking of that. 
Maybe "ls"?

wfarner:
@AkihiroSuda 

> Maybe "ls"?

Ah, yes - that's even better for the CLI command.

PR #280, opened 2016-11-07
Leadership API + polling implementations

### LINKED ISSUE #278: High availability via leadership *detection*
As mentioned in the [`README`](./README.md), InfraKit can be run in replicas with one replica as the leader and the others in hot-standby.  This HA (high availability) feature is achieved not via native implementation of leader election but instead rely on an existing system -- to the spirit of the 'toolkit'.   This means we need a mechanism/ API for third-party systems to implement a way to notify InfraKit of leadership changes.  

Common strategies for this is to use a consensus based system like Consul, Zookeeper or Docker Swarm as the source of truth on which replica runs as the leader.   In addition, we should provide a default implementation that offers zero or very low friction for test driving the HA feature.  

This is also related to the storage / state management issue #244 since same consensus based system can also be leveraged for state persistence.

chungers:
Closes #278 

  + Simple API defined for leadership detection.  See `leader.go`
  + Two implementations are included:
    
    - Simple disk file.  To be used over a shared filesystem (e.g. NFS mount), this implementation checks for the leadership by comparing the value of a file.  If a named instance (the caller) is given an id, we can determine leadership by comparing the value of the file with the name of the instance.   This is the easiest way and is good for local testing.
     - Leadership via Swarm mode -- uses Docker API / Swarm mode to decide if *this* instance is running as leader.  This assumes InfraKit replicas are running on Docker swarm manager nodes.

### DIFF
--- a/leader/file/file.go
+++ b/leader/file/file.go
@@ -4,33 +4,12 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"os/user"
-	"path/filepath"
 	"strings"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/infrakit/leader"
 )
-
-const (
-	// LeaderFileEnvVar is the environment variable that may be used to customize the plugin leader detection
-	LeaderFileEnvVar = "INFRAKIT_LEADER_FILE"
-)
-
-// DefaultLeaderFile is the file that this detector uses to decide who the leader is.
-// In a mult-host set up, it's assumed that the file system would be share (e.g. NFS mount or S3 FUSE etc.)
-func DefaultLeaderFile() string {
-	if leaderFile := os.Getenv(LeaderFileEnvVar); leaderFile != "" {
-		return leaderFile
-	}
-
-	home := os.Getenv("HOME")
-	if usr, err := user.Current(); err == nil {
-		home = usr.HomeDir
-	}
-	return filepath.Join(home, ".infrakit/leader")
-}
 
 // NewDetector return an implementation of leader detector
 // This implementation checks a file for its content.  If the content matches the id of the detector
--- a/leader/file/file.go
+++ b/leader/file/file.go
@@ -1,0 +1,59 @@
+package file
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/user"
+	"path/filepath"
+	"strings"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/docker/infrakit/leader"
+)
+
+const (
+	// LeaderFileEnvVar is the environment variable that may be used to customize the plugin leader detection
+	LeaderFileEnvVar = "INFRAKIT_LEADER_FILE"
+)
+
+// DefaultLeaderFile is the file that this detector uses to decide who the leader is.
+// In a mult-host set up, it's assumed that the file system would be share (e.g. NFS mount or S3 FUSE etc.)
+func DefaultLeaderFile() string {
+	if leaderFile := os.Getenv(LeaderFileEnvVar); leaderFile != "" {
+		return leaderFile
+	}
+
+	home := os.Getenv("HOME")
+	if usr, err := user.Current(); err == nil {
+		home = usr.HomeDir
+	}
+	return filepath.Join(home, ".infrakit/leader")
+}
+
+// NewDetector return an implementation of leader detector
+// This implementation checks a file for its content.  If the content matches the id of the detector
+// then this instance is the leader.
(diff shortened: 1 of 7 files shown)


leader/file/file_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"testing"
+	"time"
+
+	"github.com/docker/infrakit/leader"
+	"github.com/stretchr/testify/require"
+)
+
+func TestFileDetector(t *testing.T) {
+
+	file := filepath.Join(os.TempDir(), fmt.Sprintf("%d", time.Now().UnixNano()))
```
wfarner: This can be replaced with something like

```go
file := ioutil.TempFile("", "infrakit-file-test")
```

leader/leader.go
```diff
@@ -0,0 +1,35 @@
+package leader
+
+// Status indicates leadership status
+type Status int
+
+const (
+	// StatusNotLeader means the current node is not a leader
+	StatusNotLeader Status = iota
```
wfarner: Following suit with a nit you filed on `UnknownHealth` - the type is already `Status`, so should `Status` be removed from these consts?

leader/poller.go
```diff
@@ -0,0 +1,75 @@
...
+				event.Error = err
+			} else {
+				if isLeader {
+					event.Status = StatusLeader
+				} else {
+					event.Status = StatusNotLeader
+				}
+			}
+
+			l.leaderChan <- event
```
wfarner: How would you feel about making events edge-triggered instead of 'pings'?  In addition to making the system more chatty (and natural to build with an external event-based system), it will make it more apparent where to add support for something like timeouts for unknown leadership.

This should just be a matter of tracking the last event sent and suppressing if equal.

leader/poller_test.go
```diff
@@ -0,0 +1,38 @@
...
+			break
+		}
+	}
+
+	// Expect 1 event from the channel
+	event := <-events
+	require.Equal(t, StatusLeader, event.Status)
+
+	// Expect proper stopping
+	detector.Stop()
```
wfarner: After this can you do another read from `events` to ensure that it was closed?

leader/swarm/swarm_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	count := 10
+	for event := range events {
+		require.Equal(t, leader.StatusLeader, event.Status)
+		count += -1
+		if count == 0 {
+			detector.Stop()
+		}
+	}
+	// tests that we will stop cleanly here
```
wfarner: Comment seems out of place

leader/file/file.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+const (
+	// LeaderFileEnvVar is the environment variable that may be used to customize the plugin leader detection
+	LeaderFileEnvVar = "INFRAKIT_LEADER_FILE"
+)
+
+// DefaultLeaderFile is the file that this detector uses to decide who the leader is.
+// In a mult-host set up, it's assumed that the file system would be share (e.g. NFS mount or S3 FUSE etc.)
+func DefaultLeaderFile() string {
```
wfarner: This is unused, can you hold off until we have a caller?

leader/leader.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// StatusLeader means the current node / instance is a leader
+	StatusLeader
+
+	// StatusUnknown indicates some exception happened while determining leadership.  Consumer will interpret accordingly.
+	StatusUnknown
+)
+
+// CheckLeaderFunc is all that a special backend needs to implement.  It can be used with the
+// NewPoller function to return a polling implementation of the Detector interface.
```
wfarner: This doc tells more about integration than what the function is.  If i were looking to integrate, i'd want to know what true/false mean, and when i should return an `error` (i.e. what happens if i return error?)

leader/file/file_test.go
chungers: done

leader/poller.go
chungers: Edge trigger is actually implemented in the consumer of this (PR coming).  There's already a single implementation that applies a uniform handling of this 'raw' leadership data.

leader/poller_test.go
chungers: done

leader/swarm/swarm_test.go
chungers: replaced with another check on the channel to ensure it's closed.

leader/file/file.go
chungers: This is actually better in a `cmd`/ main that uses the library.  Moving to that.

leader/leader.go
chungers: Removed

PR #283, opened 2016-11-09
Manager: a stateful group plugin with leader detection

chungers:
This PR combines the earlier PRs for state storage and leader detection to create
a manager (entity + cmd binary).  The manager offers a Group interface and can record
state in a swarm or local file.   

There's a new executable `infrakit-manager` that can be started as a point of contact for users.  This binary will handle leader detection.  See `cmd/manager/README.md` for details.


manager/manager.go
```diff
@@ -0,0 +1,359 @@
...
+
+	m.Plugin = gp
+	return m, nil
+}
+
+// Start starts the manager.  It does not block. Instead read from the returned channel to block.
+func (m *manager) Start() (<-chan struct{}, error) {
+	m.lock.Lock()
+	if m.running != nil {
+		m.lock.Unlock()
```
wfarner: I'm a bit uneasy with the ad hoc locking starting to show up.

Possibly only worth dealing with later, but it would be nice to establish a straightforward convention for lock management.  It would be nice if we could strongly favor associating lock acquisition with function calls, like the pattern used elsewhere in this file:

```go
func ... {
	m.lock.Lock()
	defer m.lock.Unlock()
}
```

What do you think?

manager/manager.go
```diff
@@ -0,0 +1,359 @@
...
+}
+
+func (m *manager) onLostLeadership() error {
+	log.Infoln("Lost leadership")
+
+	// Unwatch uses a cached version in memory.  This means that we are never able to
+	// unwatch without first watching.  If the manager crashes and comes back up again,
+	// there will be no possibility to unwatch existing groups.
+	m.lock.Lock()
+	config := m.currentConfig
```
wfarner: Some functions have access to a function-scoped `GlobalSpec` as well as `m.currentConfig`, which creates potential for bugs due to not using the right spec.

Given that this is the only read of `m.currentConfig`, and the fact that you could ask the plugin for the list of groups being watched, it seems like you can get away without having `currentConfig`.

manager/manager.go
```diff
@@ -0,0 +1,359 @@
...
+
+			// TODO(chungers) -- yes this is clunky
+			if err != nil && strings.Contains(err.Error(), "Already watching") {
+
+				log.Infoln("Already WATCHING", spec.ID, "no action")
+				return nil
+			}
+
+			if err != nil {
+				log.Warningln("Error watching group:", spec.ID, "Err=", err)
```
wfarner: I suggest making this fatal, and i'd argue the same for the 'Already watching' case above to be conservative against corner cases for now.

cmd/manager/os.go
```diff
@@ -0,0 +1,81 @@
...
+	// StoreDirEnvVar is the directory where the configs are stored
+	StoreDirEnvVar = "INFRAKIT_STORE_DIR"
+)
+
+func getHome() string {
+	home := os.Getenv("HOME")
+	if usr, err := user.Current(); err == nil {
+		home = usr.HomeDir
+	}
+	return home
```
wfarner: Nit: restructure for brevity and greater readability:

```go
if usr, err := user.Current(); err == nil {
		return usr.HomeDir
}
return os.Getenv("HOME")
```

cmd/manager/main.go
```diff
@@ -0,0 +1,86 @@
...
+	snapshot store.Snapshot
+}
+
+func main() {
+
+	logLevel := cli.DefaultLogLevel
+
+	backend := &backend{
+		id: "group",
+	}
```
wfarner: I think the shared mutable state here makes the code harder to read than it should be.  Can you instead convert the `PersistentPostRunE` to a function you call and pass a `backend`?

util/docker/docker.go
```diff
@@ -0,0 +1,67 @@
+package docker
```
chungers: This code is meant to be reused (hence in a `util` package).  I am planning to change the example swarm flavor to use this instead of including a client in its own `example` package.

wfarner: Sounds good to me.  Can you do it in this PR?

chungers: Updated with versioned package for Docker client and changed the example to use the same client as a library.

cmd/manager/README.md
```diff
@@ -0,0 +1,129 @@
...
+  `group-stateless`
+  + Then when starting the manager, set the `--proxy-for-group` flag to the name of the group plugin
+  (e.g. `group-stateless`).  By default, the manager starts up with the name of `group`.  This matches
+  the default name that the CLI (`infrakit group ...`) uses.
+
+
+## Running
+
+```shell
+$ infrakit-manager -h
```
chungers: Done

cmd/manager/os.go
chungers: Done

cmd/manager/os.go
```diff
@@ -0,0 +1,81 @@
...
+		return storeDir
+	}
+	return filepath.Join(getHome(), ".infrakit/configs")
+}
+
+func osEnvironment(backend *backend) *cobra.Command {
+
+	pollInterval := 5 * time.Second
+	filename := defaultLeaderFile()
+	storeDir := defaultStoreDir()
```
chungers: I personally don't like uninitialized variables and having to look for flag bindings to know what default values are -- if the value were even bound as a flag.

In any case I will make the change assuming it's just some convention.

cmd/manager/swarm.go
```diff
@@ -0,0 +1,61 @@
...
+	cmd := &cobra.Command{
+		Use:   "swarm",
+		Short: "swarm mode for leader detection and storage",
+		RunE: func(c *cobra.Command, args []string) error {
+
+			dockerClient, err := docker.NewDockerClient(host, &tlsOptions)
+			log.Infoln("Connect to docker", host, "err=", err)
+			if err != nil {
+				log.Error(err)
+				os.Exit(1)
```
chungers: Changed

cmd/manager/os.go
wfarner: I hear you, and I don't think the design of the cobra API makes it easy to structure well.

Another problem resulting from the initializing approach here is you have two places to check for the default - the variable _and_ flag definition to ensure that it was wired into the default parameter.  I've had to correct a few places already where the latter was missing, leading to confusion.

manager/manager.go
```diff
@@ -0,0 +1,359 @@
...
+	running  chan struct{}
+
+	currentConfig GlobalSpec
+
+	backendName string
+	backendOps  chan<- backendOp
+}
+
+type backendOp struct {
+	n   string
```
chungers: Done

manager/manager.go
chungers: Changed to use an init function that returns true only when the first caller caused initialization of the channel variable.  Another concurrent caller of Start won't proceed.

manager/manager.go
```diff
@@ -0,0 +1,359 @@
...
+
+func (m *manager) doWatchGroups(config GlobalSpec) error {
+	log.Infoln("Start watching groups")
+	return m.execPlugins(config,
+		func(plugin group.Plugin, spec group.Spec) error {
+
+			log.Infoln("WATCH group", spec.ID, "with spec:", spec)
+			err := plugin.WatchGroup(spec)
+
+			// TODO(chungers) -- yes this is clunky
```
chungers: Done.  It's talking about untyped error which forces string comparison.

manager/manager.go
chungers: Changing to warning.  I really don't like to panic or have a fatal which will cause the process to exit.  In this case, if it's already watching there's no adverse effect.   Watch should be idempotent, I'd imagine.  I think logging or a warn should be enough.

cmd/manager/main.go
chungers: Done

manager/manager.go
wfarner: That's the problem - `watch` is not idempotent (it does not initiate a rolling update), so in the current behavior this signals a potential problem.

manager/manager.go
chungers: I agree that querying is better than an in-memory copy since I noted elsewhere in a comment that using the in-memory copy to notify the downstream plugins won't work if the manager exited and comes back up.  So I am changing to querying *for now* the proxied group plugin.   Note that this isn't the general case where a global config has multiple groups with different group plugins implementations running (instead of just the single one proxied here).   To construct the global spec requires querying *all* the group plugins (imagine one that's group-default and one that's backed by some autoscaling group) to construct this global collection of groups and their specs.

Anyway, added a TODO / comment and changing to querying the backend plugin for now.

manager/manager.go
wfarner: Thinking more on this, while this condition does indicate a state mismatch that the user must reconcile, making this fatal could result in a state that's difficult to reconcile.  This is because the next attempt to restart this daemon would try to watch groups, and likely fail again indefinitely.

cmd/manager/main.go
```diff
@@ -0,0 +1,84 @@
...
+	backend := &backend{}
+
+	cmd := &cobra.Command{
+		Use:   filepath.Base(os.Args[0]),
+		Short: "Manager",
+		PersistentPreRun: func(c *cobra.Command, args []string) {
+			cli.SetLogLevel(logLevel)
+		},
+		PersistentPostRunE: func(c *cobra.Command, args []string) error {
+			return runMain(backend)
```
wfarner: Incremental readability improvement on my end - eliminate the `PersistentPostRunE` and have the subcommands call `runMain()`.

manager/manager.go
chungers: My personal view on this...  if the Group's Watch operation isn't idempotent, then it's the proxied group plugin that should panic, not this layer.  This manager's concerns are only
  + persist user's spec for multiple groups
  + determine if it's the leader
  + activate plugins (including install -- TBD later)
  + update / watch on the groups

cmd/manager/main.go
chungers: Done.

PR #293, opened 2016-11-11
Collapse `watch` and `update` commands into `commit`

### LINKED ISSUE #292: Combine 'watch' and 'update' group actions
In the trend of declarative state and idempotent actions, `watch` and `update` are currently outliers.  Successfully choosing which one to invoke requires knowledge of the state of the cluster.  There should be little adverse impact to combining them, essentially treating every `watch` as an `update` (which could begin from a blank state) will make InfraKit easier to use by humans and automation.

I propose we call the new verb `commit`  meaning the following commands:

```shell
$ infrakit group watch group.json
$ infrakit group update group.json
```

would combine into
```shell
$ infrakit group commit group.json
```

The implication being that `commit` is an idempotent operation that does not depend on the current state of the cluster.

Additionally, i propose we replace the command:
```shell
$ infrakit group update-describe group.json
```
with
```shell
$ infrakit group commit group.json --pretend
```

wfarner:
Introduces the `commit` command, with a `--pretend` flag
`unwatch` is renamed to `free`
The following commands are removed: `watch`, `update`, `stop-update`, and `describe-update`

Closes #292 

### DIFF
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -260,7 +260,7 @@ func (m *manager) onLostLeadership() error {
 	if err != nil {
 		return err
 	}
-	return m.doReleaseGroups(config)
+	return m.doFreeGroups(config)
 }
 
 func (m *manager) doCommit() error {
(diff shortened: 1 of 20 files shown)


manager/manager.go
```diff
@@ -271,67 +274,36 @@ func (m *manager) doCommit() error {
...
-			if err != nil {
-				log.Warningln("Error watching group:", spec.ID, "Err=", err)
-			}
-			return nil
-		})
-}
-
-func (m *manager) doUnwatchGroups(config GlobalSpec) error {
-	log.Infoln("Unwatching groups")
+func (m *manager) doReleaseGroups(config GlobalSpec) error {
```
chungers: Just call this `doFreeGroups` to be consistent with the plugin `free` verb?

PR #294, opened 2016-11-14
Better handling on platforms where `os/user.Current()` is not implemented

### LINKED ISSUE #287: document or better handle plugin discovery directory errors on Linux
`INFRAKIT_PLUGINS_DIR`

chungers:
Closes #287 

Code should not panic when the package `os/user`'s `Current()` function isn't implemented.  Instead try to get the environment variable `HOME`. 

Changed the 'constructor' function for the directory discovery to accept a directory parameter.  This better articulates the dependency (the name of the directory) and allows future tests to better control the directory to scan for the discovery code under testing.   

Changed the mains to use the `Dir()` function which computes the default directory, as the input.

### DIFF
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -34,7 +34,12 @@ func Dir() string {
 }
 
 // NewPluginDiscovery creates a plugin discovery based on the environment configuration.
-func NewPluginDiscovery(pluginDir string) (Plugins, error) {
+func NewPluginDiscovery() (Plugins, error) {
+	return NewPluginDiscoveryWithDirectory(Dir())
+}
+
+// NewPluginDiscoveryWithDirectory creates a plugin discovery based on the directory given.
+func NewPluginDiscoveryWithDirectory(pluginDir string) (Plugins, error) {
 	stat, err := os.Stat(pluginDir)
 	if err == nil {
 		if !stat.IsDir() {
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -26,17 +26,15 @@ func Dir() string {
 		return pluginDir
 	}
 
-	usr, err := user.Current()
-	if err != nil {
-		panic(err)
+	home := os.Getenv("HOME")
+	if usr, err := user.Current(); err == nil {
+		home = usr.HomeDir
 	}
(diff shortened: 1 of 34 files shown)


discovery/discovery.go
```diff
@@ -26,17 +26,15 @@ func Dir() string {
...
+	if usr, err := user.Current(); err == nil {
+		home = usr.HomeDir
 	}
-	return path.Join(usr.HomeDir, ".infrakit/plugins")
+	return path.Join(home, ".infrakit/plugins")
 }
 
 // NewPluginDiscovery creates a plugin discovery based on the environment configuration.
-func NewPluginDiscovery() (Plugins, error) {
-	pluginDir := Dir()
```
FrenchBen: Previously the constructor grabbed the directory from the env `os.Getenv(PluginDirEnvVar)` when it needed to be different. 

I'm not sure I follow what we gain by explicitly passing it to the constructor?
```
os.Setenv("INFRAKIT_PLUGINS_DIR","/my/test/dir")
plugins, err := discovery.NewPluginDiscovery()
```
vs
```
plugins, err := discovery.NewPluginDiscovery("/my/test/dir")
```

chungers: By having the parameter in the function we are
  
  + being explicit about the requirements to construct a discovery object (the directory to scan).  Environment variable
  + following a pattern similar to IoC (inversion of control) or dependency injection -- meaning it's up to the caller to configure the object at construction time (`NewPluginDiscovery` is like a constructor in that it creates and returns a new instance).
  + some of the tests I am working on need to set this directory explicitly and the code is easier to follow with the second form, especially when the tests and os process boundaries aren't clear and tests may be executed concurrently.  

Having said that,  I am not crazy about all the changes in the main.go files.  Let me do this:

  1. Add a new function that allows setting of the directory
  2. Change the implementation of `NewPluginDiscovery() to call that new function with the `Dir()` function which returns a default location
  3. Revert the main.go files back to using the 'default constructor' form so that we have a simple, uniform default.

 @wfarner

wfarner: Yes, supporting but not requiring the arg sounds like the right thing to do.

Issue #269, opened 2016-11-01
More thorough logging needed throughout


wfarner:
From comment in #83:

> We really need to enhance the logging support. First of all, all logging should have timestamp. 2nd, the current code has little or no log, it is really hard to debug things when stuff goes wrong.

I concur with this statement that we should make a pass through major components to ensure that debug-level logging is more thorough.

linsun:
Thank you @wfarner for opening this!

wfarner:
@linsun i'm going to take a pass at this over the next few days.  If there are any particular bodies of  code where you would like to see more logging, feel free to drop a list here!

linsun:
@wfarner Thanks for asking! I would like to see log finest level for entry/exit into any method, that is helpful when troubleshooting things.  I would like to see info level logs associated to key actions, like group watch request received, group is being destroyed, group is being created. I would like to see appropriate error and warning message in the logs.  When there is something wrong, would be great to see where the error was first occurred, whether this is a user error where user can potentially correct in group.json or plugin error or infrakit code error.  I recently tried rolling upgrade, but it didn't work for me, there was not much log at all other than it did detect that I requested an increase of 1 more instance number, but it also keeps saying "Waiting for scaler to quiesce"...

wfarner:
@linsun thanks.  I agree with everything except logging every method.  However, i was looking to add full RPC request/response logging, which might be in a similar vein without as much upkeep/noise.

linsun:
@wfarner logging every method is something I am debating myself too, i think it is great to have at the finest logging level.  I totally agree with you the request/response logging would be awesome!

Issue #285, opened 2016-11-09
Run InfraKit on Windows


friism:
* when building for Windows (which works), the executable files in build should have the `.exe` extension
 * Named pipe should be supported in addition to unix socket:

```.\infrakit-group-default.exe
time="2016-11-09T07:50:20-08:00" level=info msg="Listening at: C:\\Users\\Michael Friis/.infrakit/plugins/group"
time="2016-11-09T07:50:20-08:00" level=error msg="listen unix C:\\Users\\Michael Friis/.infrakit/plugins/group: socket: An address incompatible with the requested protocol was used."
```

wfarner:
Some relevant discussion in #214

chungers:
For this it probably makes more sense to go with named pipe on Windows  https://github.com/natefinch/npipe 

TCP support #214 is related but should be focused on 'consuming' SPI providers that exposes a JSON-RPC endpoint (say a custom instance provisioning system), rather than allowing the binaries we provide to listen on TCP ports.

wfarner:
@chungers i was not suggesting restoring TCP support, just pointing to context.  In particular, this line from the issue description:

> TCP support is primarily offered as a lowest-common denominator for Windows, but we should look to named pipes for a closer analog.

PR #297, opened 2016-11-15
Use JSON-RPC over HTTP for communication between plugins

wfarner:
While looking into adding request/response logging in #269, i realized that plugin communications were using raw socket I/O rather than HTTP.  This patch uses JSON-RPC over HTTP, which i suspect will be easier for others to implement and test with conventional tools.  Specifically, this uses `github.com/gorilla/rpc`.  I chose this over `net/rpc` and `net/rpc/jsonrpc` to favor traditional HTTP request-oriented behavior.  `net/rpc` can operate over HTTP, but only as a starting point after which it hijacks connections and no longer uses HTTP requests (similar to websocket behavior).

I've elected to use `gopkg.in/tylerb/graceful.v1` for handling graceful shutdown of the HTTP server, which simplified lifecycle management.

In addition, this patch adds HTTP request logging at the debug log level.  I will follow up to refine the log output, as it currently does not log the request body.

The patch comes in 2 commits - one with the code changes, and one with vendor updates in the course of adding dependencies.


cmd/cli/flavor.go
```diff
@@ -31,10 +31,7 @@ func flavorPluginCommand(plugins func() discovery.Plugins) *cobra.Command {
 				return err
 			}
 
-			flavorPlugin, err = flavor_plugin.NewClient(endpoint.Protocol, endpoint.Address)
-			if err != nil {
-				return err
-			}
+			flavorPlugin = flavor_plugin.NewClient(endpoint.Address)
```
chungers: Why is the protocol being removed?   We need to support named pipes per #285 and this change here will have to be undone.

rpc/server/server.go
```diff
@@ -0,0 +1,68 @@
...
+	httpLog := log.New()
+	httpLog.Level = log.GetLevel()
+
+	handler := handlers.LoggingHandler(httpLog.WriterLevel(log.DebugLevel), server)
+	gracefulServer := graceful.Server{
+		Timeout: 10 * time.Second,
+		Server:  &http.Server{Addr: fmt.Sprintf("unix://%s", socketPath), Handler: handler},
+	}
+
+	listener, err := net.Listen("unix", socketPath)
```
chungers: How is this going to work for #285?  Now that we are moving this code around, we should make accommodations for the protocol string.

wfarner: I'm of the opinion that #285 will require a concerted effort that would not be mitigated by accepting a protocol string piecemeal.  I don't see much benefit to involving support for alternate protocols in this patch.

cmd/cli/flavor.go
wfarner: I'm uneasy about half-supported features.  Assuming that #285 is O(months) out before we break ground, i find the added clarity of support and simplicity welcome in the meantime.

wfarner: That said, i don't feel strongly.  So if you feel strongly about keeping the protocol plumbing on this end, i can revert.

rpc/server/server.go
chungers: It won't take O(months).   I'm planning on using https://github.com/natefinch/npipe  and it has the same convention as `net.Listen()`.  However, I will need the protocol to be passed in so that we can implement a switch.

PR #301, opened 2016-11-15
Establish a convention for declaring CLI flags

wfarner:
I'd like to propose this style as a convention for declaring CLI flags and associated variables.  I like this since it eliminates the data flow cycle we frequently had otherwise (declare a zeroed var, refer to zeroed var for flag default, refer to var pointer for arg value).

### DIFF
--- a/cmd/cli/flavor.go
+++ b/cmd/cli/flavor.go
@@ -18,25 +18,25 @@ import (
 
 func flavorPluginCommand(plugins func() discovery.Plugins) *cobra.Command {
 
-	name := ""
 	var flavorPlugin flavor.Plugin
 
 	cmd := &cobra.Command{
 		Use:   "flavor",
 		Short: "Access flavor plugin",
-		PersistentPreRunE: func(c *cobra.Command, args []string) error {
-
-			endpoint, err := plugins().Find(name)
-			if err != nil {
-				return err
-			}
-
-			flavorPlugin = flavor_plugin.NewClient(endpoint.Address)
-
-			return nil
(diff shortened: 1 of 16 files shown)


wfarner:
If we achieve consensus on this style, i will go ahead and update the rest of the codebase in this PR.

cmd/cli/flavor.go
```diff
@@ -18,25 +18,25 @@ import (
...
 
-	name := ""
 	var flavorPlugin flavor.Plugin
 
 	cmd := &cobra.Command{
 		Use:   "flavor",
 		Short: "Access flavor plugin",
-		PersistentPreRunE: func(c *cobra.Command, args []string) error {
+	}
+	name := cmd.PersistentFlags().String("name", "", "Name of plugin")
```
FrenchBen: Does this really need to be a `PersistentFlags()`? 
It doesn't seem that the name is passed down to any sub-children, is it?

wfarner: It does need to be, as you want to specify a plugin `name` for most of the subcommands being used.  IMHO it really should be a positional (required) argument rather than a flag, but that's not something i'm willing to tackle in this patch.

chungers:
I think having pointer deferences in the function bodies look a bit awkward.

That said, if it's just a convention / pattern then I am ok with it.

wfarner:
@chungers i agree, but i see it as a lesser of evils.  IMHO this pattern reduces likelihood of subtle bugs, for a net improvement.

PR #302, opened 2016-11-16
Run tutorial test in `make ci`

wfarner:


wfarner:
```shell
ERRO[0000] listen unix /home/ubuntu/.infrakit/plugins/instance-file: bind: no such file or directory 
ERRO[0000] listen unix /home/ubuntu/.infrakit/plugins/flavor-vanilla: bind: no such file or directory 
```

I'll need to sort this out.  Could be a permissions issue or some behavior difference between Darwin and Linux.

PR #306, opened 2016-11-17
Add tests for the Manager plugin

### LINKED ISSUE #298: Add tests for the manager plugin
Follow up from #283 by adding unit test coverage.

chungers:
Closes #298 

Changes

  + New mocks
  + New tests to verify behavior of two ensembles -- each with its own manager and group plugins -- for the cases
     - neither one of the ensembles is the leader
     - one of the ensembles become leader
     - one ensemble was leader then leadership is transferred / fail-overed to another

There are a bunch of unsigned merge commits which Gordon seems to get confused and complains about.  We use squash merge so these should be ok.


pkg/manager/manager_test.go
```diff
@@ -0,0 +1,265 @@
...
+	"time"
+
+	"github.com/docker/infrakit/pkg/discovery"
+	leader_file "github.com/docker/infrakit/pkg/leader/file"
+	group_mock "github.com/docker/infrakit/pkg/mock/spi/group"
+	store_mock "github.com/docker/infrakit/pkg/mock/store"
+	group_rpc "github.com/docker/infrakit/pkg/rpc/group"
+	"github.com/docker/infrakit/pkg/rpc/server"
+	"github.com/docker/infrakit/pkg/spi/group"
+	_ "github.com/docker/infrakit/pkg/store"
```
wfarner: What's the purpose of the underscore import?  If it remains, can you leave a comment so it's clear why it needs to be here?

pkg/manager/manager_test.go
```diff
@@ -0,0 +1,265 @@
...
+			// no calls
+		},
+		func(g *group_mock.MockPlugin) {
+			// no calls
+		})
+
+	manager1.Start()
+	manager2.Start()
+
+	time.Sleep(1 * time.Second)
```
wfarner: As discussed offline, if you use a custom leader detector for test, you should be able to eliminate (or at least more tightly control) the leadership delay.  A nice outcome for this will be to remove all the calls to `time.Sleep()`.

pkg/manager/manager_test.go
chungers: removed anonymous import

pkg/manager/manager_test.go
chungers: see updates

pkg/manager/manager_test.go
```diff
@@ -0,0 +1,312 @@
...
+func (l *testLeaderDetector) Stop() {
+	close(l.stop)
+}
+
+func testEnsemble(t *testing.T,
+	dir, id string,
+	leader chan string,
+	ctrl *gomock.Controller,
+	configStore func(*store_mock.MockSnapshot),
+	configureGroup func(*group_mock.MockPlugin)) (Backend, server.Stoppable) {
```
FrenchBen: I'm not a big fan of function that starts to take on a ton of parameters. 
Anyway for us to clean this up via struct or split this until smaller functions?

Perfect example in Kubernetes:
https://github.com/kubernetes/kubernetes/blob/973f2fcd86ef3c3e69f455aaf88553f0906b48bd/pkg/kubelet/kubelet.go#L171

Now refactored to:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet.go#L272

chungers: I think you're referring to the long list of parameters in the function `NewManager` in `manager.go`?

This function uses parameters that are of complex type (e.g. `discoverry.Plugin`, `leader.Detector`) instead of simple types like strings and bools (seen in the k8s you referenced), so while the function arg list is long, it's hard to mess things up without the compiler complaining.  

I could look into doing something with embedding / composition with a struct like a `manager.Config`; however, my concern is that the fields would be public/exported and can be mutated after a manager instance is initialized.  Having the dependencies passed in as function params allow the manager's dependencies (as struct fields) to be unexported and immutable after initialization.

pkg/manager/manager_test.go
```diff
@@ -234,32 +275,38 @@ func TestChangeLeadership(t *testing.T) {
...
+				},
+			).Return(nil)
 		},
 		func(g *group_mock.MockPlugin) {
 			g.EXPECT().CommitGroup(gomock.Any(), false).Do(
 				func(spec group.Spec, pretend bool) (string, error) {
+
+					defer close(checkpoint2)
+
+					t.Log("m2 leading")
```
wfarner: Remove debug log statement

pkg/manager/manager_test.go
```diff
@@ -208,24 +236,37 @@ func TestChangeLeadership(t *testing.T) {
...
 			g.EXPECT().InspectGroups().Return([]group.Spec{gs}, nil)
 
 			// Now we lost leadership... need to unwatch
-			g.EXPECT().FreeGroup(gomock.Eq(group.ID("managers"))).Return(nil)
+			g.EXPECT().FreeGroup(gomock.Eq(group.ID("managers"))).Do(
+				func(id group.ID) error {
+
+					defer close(checkpoint3)
 
+					t.Log("m1 releasing")
```
wfarner: Remove debug log statement

PR #316, opened 2016-12-01
Add a Type field to Attachments

### LINKED ISSUE #312: Include types on Instance Attachments
Currently the Instance plugin api has the notion of attachments but Attachment is just a string.  This makes it hard to express or determine the type of resource attachments one can make to an instance during `Provision`. 

The current use case for attachment is for AWS EC2 instance plugin to attach a vm instance to an EBS volume.   However, there are other resource types where the attachment semantic fits.  For example, AWS just announced Elastic GPU that allows GPU to be attached to an instance. 

The change proposed here is to simply make `Attachment` a complex type of two fields:
```
{
   "type" : "ebs",
   "id" : "ebs-instance-id"
}
``` 
An EC2 instance can potentially support attaching an EBS volume and EGPU:
```
{
    "Attachments" : [ { "type":"ebs", "id" : "ebs-123"}, { "type" : "egpu", "id", "gpu-1234" } ]
}
```

To keep things simple, the `type` is just a string that the instance plugin understands.  This is implicit, but doing cross-plugins data validation is too much to take on at the moment.

wfarner:
Closes #312 

### DIFF
--- a/docs/plugins/instance.md
+++ b/docs/plugins/instance.md
@@ -1,6 +1,6 @@
 # Instance plugin API
 
-<!-- SOURCE-CHECKSUM pkg/spi/instance/* 0c778e96cbeb32043532709412e15e6cc86778d7153c65b1d7422b9f0b5901c89362666bfdd75de7 -->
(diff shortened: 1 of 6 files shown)


chungers:
You beat me to it!!  Thanks for this PR!

PR #318, opened 2016-12-03
Add Version information to plugin APIs

### LINKED ISSUE #309: Include versioning information in plugin APIs
We are about to publish two infrakit third party plugins. We were wondering how to tie our plugins to a specific release or revision?

To be more specific for example recently you changed `infrakit group watch` to `commit`. Having a plugin tutorial might look incorrect after a change.  

Also since plugins implement `instance.Plugin` do you plan to change the interface sometimes in the future?

wfarner:
Publishing just to share the approach taken so far.  Feedback welcome.

Closes #309

### DIFF
--- a/pkg/rpc/client/handshake.go
+++ b/pkg/rpc/client/handshake.go
@@ -1,0 +1,66 @@
+package client
+
+import (
+	"fmt"
+	"github.com/docker/infrakit/pkg/rpc/plugin"
+	"github.com/docker/infrakit/pkg/spi"
+	"sync"
+)
+
+type handshakingClient struct {
+	client Client
+	api    spi.APISpec
+
+	handshakeResult *handshakeResult
+
+	// lock guards handshakeComplete
+	lock *sync.Mutex
+}
+
+type handshakeResult struct {
+	err error
+}
+
+func (c *handshakingClient) handshake() error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	if c.handshakeResult == nil {
+		req := plugin.APIsRequest{}
+		resp := plugin.APIsResponse{}
+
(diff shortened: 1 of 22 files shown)


pkg/rpc/client/handshake.go
```diff
@@ -0,0 +1,66 @@
...
+	"sync"
+)
+
+type handshakingClient struct {
+	client Client
+	api    spi.APISpec
+
+	handshakeResult *handshakeResult
+
+	// lock guards handshakeComplete
```
chungers: The comment - do you mean `handshakeResult`?

wfarner: Yep, thanks.  Fixing.

pkg/spi/plugin.go
```diff
@@ -0,0 +1,16 @@
+package spi
```
chungers: As a nit -- would you mind rename `APISpec` to something like `Interface`?    `API` looks jarring when there are names like `APIsRequest` and `APIsResponse`.  

Also for the method `APIs()`, `Implements()` looks a little nicer to me.  It also matches how Docker engine does activation with the field `Implements`:  https://docs.docker.com/engine/extend/plugin_api/#/pluginactivate

So how about 
```
type Interface struct {
   Name string
   Version string
}

type Plugin interface {
    Implements() ([]Interface, error)
}
```
Then the RPC response can be named `HandshakeResponse` and it will look like
```
{
     "Implements" : [ ....]
     "Interfaces" : [  /* list of interface descriptions like example input / output via reflection */ ]
}
```

wfarner: No strong opinions here, the names you gave sound fine to me.

PR #319, opened 2016-12-04
Self-documenting metadata / version api with example RPC input/output via reflection

chungers:
For #309 

This PR attempts to provide a self-documenting RPC endpoint to expose the following:
  1. The name / version of the infrakit plugin (e.g. Instance)
  2. The name / version of the custom plugin implementation that conforms to 1.  For example, "aws-instance", version "1.0" which implements the Infrakit "Instance" api at version "0.1"
  3. An additional field call `Interfaces` in the response of the metadata call that contains the interface methods and corresponding example request/response pairs.  This allows the user to take the example requests and with slight modification call the actual RPC methods as input (e.g. via curl).  This hopefully can provide as much documentation as possible of the API without using tools like swagger (which is more for REST api than for JSON-RPC)
  4. Support for exposing a plugin's custom type.  For example, an AWS instance plugin that uses some SDK custom type/struct can now return an example of that data and have that in the api documentation too.

Example output:

```
~/projects/src/github.com/docker/infrakit$ curl -s --unix-socket ~/.infrakit/plugins/instance-file  \
-X POST -d '{ "jsonrpc" : "2.0", "method":"Plugin.Meta", "id":"test", "params":[] }' \
-H "Content-Type: application/json"  http://d/       | jq .
{
  "result": {
    "Vendor": {
      "Name": "infrakit-instance-file",
      "Version": "v0.1.0"
    },
    "Implements": [
      {
        "Name": "Instance",
        "Version": "v0.1.0"
      }
    ],
    "Interfaces": [
      {
        "Name": "Instance",
        "Version": "v0.1.0",
        "Methods": [
          {
            "Request": {
              "jsonrpc": "2.0",
              "method": "Instance.DescribeInstances",
              "params": {
                "Tags": null
              },
              "id": "1480892865"
            },
            "Response": {
              "result": {
                "Descriptions": null
              },
              "id": "1480892865"
            }
          },
          {
            "Request": {
              "jsonrpc": "2.0",
              "method": "Instance.Destroy",
              "params": {
                "Instance": ""
              },
              "id": "1480892865"
            },
            "Response": {
              "result": {
                "OK": false
              },
              "id": "1480892865"
            }
          },
          {
            "Request": {
              "jsonrpc": "2.0",
              "method": "Instance.Provision",
              "params": {
                "Spec": {
                  "Properties": {
                    "exampleBool": true,
                    "exampleInt": 1,
                    "exampleString": "a_string"
                  },
                  "Tags": null,
                  "Init": "",
                  "LogicalID": null,
                  "Attachments": null
                }
              },
              "id": "1480892865"
            },
            "Response": {
              "result": {
                "ID": null
              },
              "id": "1480892865"
            }
          },
          {
            "Request": {
              "jsonrpc": "2.0",
              "method": "Instance.Validate",
              "params": {
                "Properties": {
                  "exampleBool": true,
                  "exampleInt": 1,
                  "exampleString": "a_string"
                }
              },
              "id": "1480892865"
            },
            "Response": {
              "result": {
                "OK": false
              },
              "id": "1480892865"
            }
          }
        ]
      }
    ]
  },
  "error": null,
  "id": "test"
}
```
Note there are blocks like
```
                  "Properties": {
                    "exampleBool": true,
                    "exampleInt": 1,
                    "exampleString": "a_string"
                  },
```
This corresponds to the custom data defined within the instance plugin itself.  In this case it's defined in file instance's [plugin.go](https://github.com/docker/infrakit/pull/319/files#diff-00600859da4a1e94e7985d9cd5094b1fR59).


wfarner:
This overlaps with #318 posted a few days ago, which focuses on solely the API layer.  How about we combine efforts?

chungers:
Yes I noticed the other PR and I think we should combine the api versioning and reflection / example input/output.

chungers:
Let's get PR #318 in and then I can merge in the reflection code.  For that PR, I've suggested response schema to hopefully make the combining easier to do.

chungers:
Slight change after PR #318 was merged:

Since all RPC calls require a version check, this endpoint is implemented as standard HTTP GET at a well-known endpoint of `/metaz` so that any tools can inspect a plugin's interfaces.

This simplifies the retrieval of the api information:

```
$ curl -s --unix-socket ~/.infrakit/plugins/instance-file  http://d/metaz | jq .
{
  "Vendor": {
    "Name": "infrakit-instance-file",
    "Version": "v0.1.0"
  },
  "Implements": [
    {
      "Name": "Instance",
      "Version": "0.1.0"
    }
  ],
  "Interfaces": [
    {
      "Name": "Instance",
      "Version": "0.1.0",
      "Methods": [
        {
          "Request": {
            "jsonrpc": "2.0",
            "method": "Instance.DescribeInstances",
            "params": {
              "Tags": null
            },
            "id": "1481070566"
          },
          "Response": {
            "result": {
              "Descriptions": null
            },
            "id": "1481070566"
          }
        },
        {
          "Request": {
            "jsonrpc": "2.0",
            "method": "Instance.Destroy",
            "params": {
              "Instance": ""
            },
            "id": "1481070566"
          },
          "Response": {
            "result": {
              "OK": false
            },
            "id": "1481070566"
          }
        },
        {
          "Request": {
            "jsonrpc": "2.0",
            "method": "Instance.Provision",
            "params": {
              "Spec": {
                "Properties": {
                  "exampleBool": true,
                  "exampleInt": 1,
                  "exampleString": "a_string"
                },
                "Tags": null,
                "Init": "",
                "LogicalID": null,
                "Attachments": null
              }
            },
            "id": "1481070566"
          },
          "Response": {
            "result": {
              "ID": null
            },
            "id": "1481070566"
          }
        },
        {
          "Request": {
            "jsonrpc": "2.0",
            "method": "Instance.Validate",
            "params": {
              "Properties": {
                "exampleBool": true,
                "exampleInt": 1,
                "exampleString": "a_string"
              }
            },
            "id": "1481070566"
          },
          "Response": {
            "result": {
              "OK": false
            },
            "id": "1481070566"
          }
        }
      ]
    }
  ]
}
```

pkg/rpc/client/client.go
```diff
@@ -2,13 +2,14 @@ package client
...
-	"github.com/docker/infrakit/pkg/spi"
-	"github.com/gorilla/rpc/v2/json2"
 	"net"
 	"net/http"
 	"net/http/httputil"
 	"sync"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/docker/infrakit/pkg/spi"
+	"github.com/gorilla/rpc/v2/json2"
```
wfarner: Revert?  If we're going to churn on import convention, i'd very much prefer we have a tool mandate the convention.

wfarner:
I'm a -1 to the one-off HTTP endpoint.

> Since all RPC calls require a version check...so that any tools can inspect a plugin's interfaces

I'm not seeing the problem that warrants protocol bifurcation.  Why are we worried about other tools with this call in particular?

chungers:
The reasoning here is that it's really not part of the API itself, but rather as a simple discovery mechanism of the interfaces that are exposed by the plugins, with example input/output.

A parallel of this would be `/varz` which exposes parameters / metrics in the server while the server itself wouldn't have a RPC method to get parameters.

Also, by exposing this as itself a RPC call comes with some ease of use challenges -- the user has to be able to properly construct a JSON-RPC call to be able to get the interfaces and methods.  Using a simple endpoint (and not a *protocol change* since it's still HTTP) is a lot easier as shown by the simpler curl command above.

wfarner:
> A parallel of this would be /varz which exposes parameters / metrics in the server while the server itself wouldn't have a RPC method to get parameters.

Not a good analogy IMHO, since /varz applied to a diverse environment of applications that used many protocols.
Also, can't the same be said of the `Interfaces` call?  Why didn't we do that as a one-off?

> Also, by exposing this as itself a RPC call comes with some ease of use challenges -- the user has to be able to properly construct a JSON-RPC call to be able to get the interfaces and methods

The same can be said of the entire API, right?  Why is this different?  Perhaps it's because you're thinking of somebody consuming the raw JSON, when in fact we might benefit from an enhanced presentation in the `meta` subcommand?

> not a protocol change since it's still HTTP

By that logic, we could switch to SOAP and call it the same protocol.  Or sockets since it's still TCP :-)

thebsdbox:
JSON-RPC doesn't recommend HTTP GET so you may hit upon implementations of HTTPD/JSON-RPC where they only support POST. 

Alternatively you could support `$ ./infrakit-instance-file --implementation` (perhaps pipe it to a file to edit etc..)

chungers:
I originally had this as a separate RPC method but thought that it was cumbersome for users to have to construct a JSON RPC request payload and set headers in curl just to find out what other RPC calls look like.  Feels kinda circular ;)  This definitely came up when you look at how handshake was implemented in #318 -- the rpc client itself requires an interface spec before it can make a RPC call -- this means adding a `meta` verb in the CLI would require working around that because the whole point of this was to expose what the plugin actually implements.

Another motivation for a separate simpler HTTP GET endpoint is that it mimics how some swagger servers in making the api definition available at a well known `/api.json` url that's out of band of the actual API that a server provides.  Similarly, WSDL w.r.t. SOAP.

In any case, I can expose all this as a new field in [`ImplementsResponse`](https://github.com/docker/infrakit/blob/master/pkg/rpc/plugin/types.go#L10) and it'd be less work.  However that comes with the tradeoff of additional overhead in the handshake (larger payload).

Another option is to do as @thebsdbox suggests -- have the binary dump out the data with a flag.  That wouldn't help with plugins that are run as remote services, though (which we would like to support one day).

wfarner:
> ...users to have to construct a JSON RPC request payload and set headers in curl just to find out what other RPC calls look like

Does it make sense to think of it in terms of something like the `meta` command you introduced rather than curl?  With that, we can provide more user-friendly output than raw JSON anyway.

chungers:
I think having the `meta` command is a definite yes.  It will help a lot.  

I am also thinking of the case where the cli binary isn't readily available and the operator just wants to a quick a easy way to do sanity check / troubleshoot.  curl seems pretty ubiquitous.

wfarner:
Seems like an odd scenario...i thought this endpoint was designed for plugin developers?

I'm still feeling like we shouldn't over-optimize and stick with the protocol, build good presentation under `infrakit plugin meta`.  If that turns out to not be the right approach we can revisit.

chungers:
Updated and simplified significantly.  Now a cli command called `Info` returns information about the plugin.  The user can point at any plugin and the cli will give details about what the plugin implements, including vendor-specific information, if the plugin uses the golang library and implement some optional interfaces.

```shell
$ build/infrakit info --name instance-file
{
  "Vendor": {
    "Name": "infrakit-instance-file",
    "Version": "0.1.0",
    "URL": "https://github.com/docker/infrakit"
  },
  "Implements": [
    {
      "Name": "Instance",
      "Version": "0.1.0"
    }
  ],
  "Interfaces": [
    {
      "Name": "Instance",
      "Version": "0.1.0",
      "Methods": [
        {
          "Request": {
            "jsonrpc": "2.0",
            "method": "Instance.DescribeInstances",
            "params": {
              "Tags": null
            },
            "id": "1482655070"
          },
          "Response": {
            "result": {
              "Descriptions": null
            },
            "id": "1482655070"
          }
        },
        {
          "Request": {
            "jsonrpc": "2.0",
            "method": "Instance.Destroy",
            "params": {
              "Instance": ""
            },
            "id": "1482655070"
          },
          "Response": {
            "result": {
              "OK": false
            },
            "id": "1482655070"
          }
        },
        {
          "Request": {
            "jsonrpc": "2.0",
            "method": "Instance.Provision",
            "params": {
              "Spec": {
                "Attachments": null,
                "Init": "",
                "LogicalID": null,
                "Properties": {
                  "exampleBool": true,
                  "exampleInt": 1,
                  "exampleString": "a_string"
                },
                "Tags": null
              }
            },
            "id": "1482655070"
          },
          "Response": {
            "result": {
              "ID": null
            },
            "id": "1482655070"
          }
        },
        {
          "Request": {
            "jsonrpc": "2.0",
            "method": "Instance.Validate",
            "params": {
              "Properties": {
                "exampleBool": true,
                "exampleInt": 1,
                "exampleString": "a_string"
              }
            },
            "id": "1482655070"
          },
          "Response": {
            "result": {
              "OK": false
            },
            "id": "1482655070"
          }
        }
      ]
    }
  ]
}
```

PR #323, opened 2016-12-06
add test code for save load swarm snapshot

YujiOshima:
Test code for save, load swarm snapshot in pkg/store/swarm/swarm.go.


wfarner:
Thanks for the patch, great stuff!

Issue #308, opened 2016-11-22
InfraKit panic when directory tree isn't present


thebsdbox:
If the `pwd`/.infrakit/plugins/ location for plugins to create their UNIX socket file does not exist, then the plugin will panic. Example below, the plugins need to create the directory structure should it not exist, or exit safely with a warning message. 

Having the plugins create the directory would allow more control over over permissions of the .infrakit directory.

```
dan@localhost ~ $ ./infrakit-instance-file &
[1] 2615
dan@localhost ~ $ ERRO[0000] listen unix /home/dan/.infrakit/plugins/instance-file: bind: no such file or directory 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x46a122]

goroutine 1 [running]:
panic(0x7276e0, 0xc42000c090)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/docker/infrakit/pkg/cli.RunPlugin(0x784f99, 0xd, 0x739440, 0xc42000d2a0)
	/root/go/src/github.com/docker/infrakit/pkg/cli/serverutil.go:18 +0x102
main.main.func1(0xc42007cb40, 0x94dfd8, 0x0, 0x0)
	/root/go/src/github.com/docker/infrakit/pkg/example/instance/file/main.go:23 +0xd1
github.com/docker/infrakit/vendor/github.com/spf13/cobra.(*Command).execute(0xc42007cb40, 0xc42000c230, 0x0, 0x0, 0xc42007cb40, 0xc42000c230)
	/root/go/src/github.com/docker/infrakit/vendor/github.com/spf13/cobra/command.go:636 +0x443
github.com/docker/infrakit/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc42007cb40, 0xc420045f28, 0x77f8e0, 0xc42007cd80)
	/root/go/src/github.com/docker/infrakit/vendor/github.com/spf13/cobra/command.go:722 +0x367
github.com/docker/infrakit/vendor/github.com/spf13/cobra.(*Command).Execute(0xc42007cb40, 0xc420045f20, 0x1)
	/root/go/src/github.com/docker/infrakit/vendor/github.com/spf13/cobra/command.go:681 +0x2b
main.main()
	/root/go/src/github.com/docker/infrakit/pkg/example/instance/file/main.go:28 +0x2cc

[1]+  Exit 2                  ./infrakit-instance-file
[...]

PR #329, opened 2016-12-09
Ensure plugin discovery directory is in place 

chungers:
Small fixes for #308 

  + Ensures the discovery directory is present
  + Don't panic when server can't bind on start up.

This doesn't fix the issue of orphaned socket files after plugins crashed.  That will be in a future PR.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -3,7 +3,7 @@ machine:
     OS: "linux"
     ARCH: "amd64"
 
-    GOVERSION: "1.6"
+    GOVERSION: "1.7"
     GOPATH: "$HOME/.go_workspace"
 
     WORKDIR: "$GOPATH/src/github.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME"
--- a/pkg/cli/serverutil.go
+++ b/pkg/cli/serverutil.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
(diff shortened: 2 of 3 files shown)


pkg/discovery/discovery.go
```diff
@@ -9,6 +9,12 @@ import (
 	"github.com/docker/infrakit/pkg/plugin"
 )
 
+// Do this once, globally to ensure the discovery directory exists
+func init() {
```
wfarner: -1.  Importing a package shouldn't create a directory.

chungers: If you're importing a package that does discovery based on the presence of a directory, why is creating it a bad thing?

chungers: If you're importing a package that does discovery based on the presence of a directory, why is creating it a bad thing?

chungers: If you're importing a package that does discovery based on the presence of a directory, why is creating it a bad thing?

chungers:
As a point of interest...  For Go the role of package init function in setting up proper state before program execution:  https://golang.org/doc/effective_go.html#init
Consequently it's possible to import packages purely for side effects: 
https://golang.org/doc/effective_go.html#blank_import

As I mentioned in the comment for #308  -- this PR is pretty much a stop-gap until a more effective plugin management strategy is implemented.

Issue #331, opened 2016-12-11
[Question] Using autoscaling API with infrakit.


anarcher:
Hi, 

I'm looking for Infrakit for using our own instances management.  If I have to create 1000 >= instances with Infrakit,  `infrakit-instance` plugin makes 1000 API Requests. Because GCE and AWS have API rate limit,All API requests are forced to take time and restart to succeed.  I think it better to use clouders' autoscaling API..  What do you think about using Infrakit for large scale instance management?   According to current plugin design, `Instance.Provisioin` means that 1 instance provision. 

Thanks,
MS

YujiOshima:
Hi @anarcher !
My opinion, `Instance.Provisioin` means only 1 instance provision make responsibility of plugin clear.
So `Aws-AutoScaleGroup-instance-plugin` or auto scale configuration in terraform layer is better to solve API rate problem.

anarcher:
Hi,@YujiOshima

Thanks for reply. :-)   With current instance plugin interface, `AWS-ASG-Instance-plugin` should has a buffer or queue of `provision` request for reducing AWS API request.(e.g. After some duration, Set DesiredSize) Is it right?  

Thanks,
MS

chungers:
@anarcher @YujiOshima 

1. The semantics of `Instance.Provision` is to provision one instance at a time.  Using an instance plugin that wraps an AWS ASG (autoscaling group) would be confusing because there will now be two controllers -- the running InfraKit default group plugin that uses this instance plugin and the platform's scaling group controller.

2. It's possible to provide an alternate implementation of the Group plugin backed by an AWS ASG.  This however would have some differences in how volumes are managed.  There isn't a reference implementation yet, but we are thinking of creating one for AWS ASG so any help / interests in this are welcome.

3.  Because we'd like to keep writing Instance plugins simple, buffering/queueing  `Instance.Provision` calls and handling retries for retry-able errors really should be implemented in the default Group plugin.  Currently the [scaler code](https://github.com/docker/infrakit/blob/master/pkg/plugin/group/scaler.go#L234) doesn't take these conditions into account, but adding proper handling of errors and retries there will hopefully solve the problem for all platforms and keep the instance plugin implementations simple.

Thoughts?

anarcher:
Thanks for the detailed explanation @chungers :-)

In my experience,when managing an instance with ASG, I did not have any complicated management of the volume (if I needed volumes in it, I simply made the volume life cycle equal to the instance life cycle).

[...]

PR #332, opened 2016-12-12
add buffering/queueing function in default Group plugin

YujiOshima:
Related #331 
Add max-parallel flag to  set maximum number of parallel provision.


YujiOshima:
CI error will fix by #333

pkg/plugin/group/scaler.go
```diff
@@ -242,6 +259,10 @@ func (s *scaler) converge() {
 
 				s.scaled.CreateOne(nil)
 			}()
+			if s.maxParallelNum > 0 && (i+1)%int(s.maxParallelNum) == 0 {
```
chungers: How about extract this into a function on the scaler:

```
func (s *scaler) waitIfReachParallelLimit(current int, batch sync.WaitGroup) {
   // call wait on the batch if the current index reaches the parallel limit
}
```
and then call this also after line [248](https://github.com/docker/infrakit/pull/332/files#diff-fdb002f7e382abf7f4b1a8a761139830R248) so that we can rate limit the Destroy calls as well?

pkg/plugin/group/scaler_test.go
```diff
@@ -13,7 +13,7 @@ func TestScaleUp(t *testing.T) {
 	defer ctrl.Finish()
 
 	scaled := mock_group.NewMockScaled(ctrl)
-	scaler := NewScalingGroup(scaled, 3, 1*time.Millisecond)
+	scaler := NewScalingGroup(scaled, 3, 1*time.Millisecond, 0)
```
chungers: Can we add a new test or test this with `maxParallel = 1`?   Maybe change the case to create more than one instance.  The test should pass fine since we are effectively serializing the CreateOne calls.

YujiOshima:
@chungers Thank you for your review!
I updated following the comments and rebase #329.

PR #336, opened 2016-12-13
Update vendored docker pkg

### LINKED ISSUE #335: Swarm flavor tutorial failed with docker api error.
I got this error when I try to provision worker node in swarm flavor tutorial.

`Failed to Prepare instance: Failed to fetch node self info: Error reading remote info: json: cannot unmarshal string into Go value of type types.SecurityOpt`

It is because `docker info` api changer in v1.25.
See https://github.com/docker/docker/pull/28504#issuecomment-261384755

It needs to update vendored docker version.

YujiOshima:
Fix #335 


chungers:
Thanks for this PR.  I am hesitant to merge this in because the Docker 1.13 release is not final / GA yet.  

Since we can build in containers, we can build conditionally.  This means leaving the vendor sources at the current stable version of Docker but provide Makefile targets to build against a newer version of the api.   The makefile target will build in a container with  (and also changing the `1.24` client version string)

chungers:
Thank you @YujiOshima for your hard work.

I just merged #338  so I think this PR is no longer needed.  Right?

YujiOshima:
@chungers Yes, I will close. Thank you.

PR #337, opened 2016-12-13
Add trash vendor tool as part of tools install

FrenchBen:
Since [trash](https://github.com/rancher/trash) is being used when updating vendoring, it'd be nice to install it as part of the tools install. 

cc @chungers @wfarner


chungers:
Thanks!

PR #338, opened 2016-12-14
Tools for building with different versions of Docker api

chungers:
This PR addresses #335   This PR does not update the vendored sources to the Docker 1.13.  Instead, it

  + Allows a different version of docker api (or any dependencies for that matter) be build inside
a container.  Simply update the `vendor.conf` with the SHA of the packages you want and run
`make build-in-container`.  This will result in a slower build in the container but the container will build with the versions of vendored dependencies you want.  Also, it will not mess up your current workspace.
  + Moves the docker client package so that no version string (`1.24`) is in the package name.  If we have code that is version specific, we will explore using build tags to allow conditional builds.
  + It's possible to set the client version string when building: 

```
DOCKER_CLIENT_VERSION=1.25 make binaries
``` 
   otherwise, 1.24 is the default.  Be sure to have the `vendor.conf` reflect the new version to use in vendor packages.
  + Prints out a Docker version string for sanity checks.


YujiOshima:
@chungers Infrakit using docker client commit id ccba791 (https://github.com/docker/infrakit/blob/master/vendor.conf#L14).
This commit is Docker 1.13 rc and API version 1.25.
https://github.com/docker/docker/blob/ccba7911c80583c1a45ea329a8d337dcd4f6f70e/api/swagger.yaml#L14
This is not compatible with 1.24 in Info command even if set `1.24` in Client.version.
So swarm-flavor still fail to provision docker v1.12 and earlier.

As Docker API change drastically from 1.12.x to 1.13.x, it is difficult to use v1.12.4 client.
It is better update vendored docker client at least 3a68292 or later.

Current status that swarm-flavor tutorial doesn't is so inconvenient. 

Thought?

chungers:
@YujiOshima  - how about I update the vendor.conf to use the version at 3a68292?

This way we should be up to date with stable version at 1.24.  If someone wants to build with 1.13 rc api (1.25), the methods introduced in this PR can be used:

  1. Update the vendor.conf to use the SHA you want
  2. Run `DOCKER_CLIENT_VERSION=$VERSION make build-in-container`

where $VERSION should match whatever version of docker API you use in vendor.conf.

When this is done, the binaries will be in build/ as usual.   How does this sound?

YujiOshima:
@chungers Sounds great! Thank you for your kind support!

PR #339, opened 2016-12-15
Automate building of Docker images

chungers:
This adds automation of building docker containers to ease development.  The build `infrakit/bundle` image is for development only and not meant for production deployment.


chungers:
I need to merge this in so that circle can build images on master.

PR #347, opened 2017-01-05
Simplify Dockerfiles for build and run

dgageot:
@chungers Is there some kind of integration test to make sure it still works?

Signed-off-by: David Gageot <david@gageot.net>

### DIFF
--- a/dockerfiles/Dockerfile.build
+++ b/dockerfiles/Dockerfile.build
@@ -1,25 +1,11 @@
 FROM golang:1.7.4-alpine
 
-# A container for building InfraKit
-
 RUN apk add --update git make
+RUN go get github.com/rancher/trash
 
-ENV CGO_ENABLED=0
-ENV GOPATH=/go
-ENV PATH=/go/bin:/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
-
-# Development tools
-RUN go get github.com/rancher/trash \
-           github.com/golang/lint/golint \
-           github.com/golang/mock/gomock \
-           github.com/golang/mock/mockgen
-
-# The project sources
-ADD ./ /go/src/github.com/docker/infrakit
 WORKDIR /go/src/github.com/docker/infrakit
-
 VOLUME [ "/go/src/github.com/docker/infrakit/build" ]
-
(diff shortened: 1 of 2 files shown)


dockerfiles/Dockerfile.build
```diff
@@ -1,25 +1,11 @@
...
-           github.com/golang/lint/golint \
-           github.com/golang/mock/gomock \
-           github.com/golang/mock/mockgen
-
-# The project sources
-ADD ./ /go/src/github.com/docker/infrakit
 WORKDIR /go/src/github.com/docker/infrakit
-
 VOLUME [ "/go/src/github.com/docker/infrakit/build" ]
+CMD make build-binaries
```
chungers: Moving the CMD at the end of the file to here is ok, right?  CMD isn't run so this is just setting the default command.   I've seen many Dockerfiles where ENTRYPOINT and CMD come at the end of the file so let me know if this is a better convention.

dockerfiles/Dockerfile.build
```diff
@@ -1,25 +1,11 @@
...
 
+COPY . ./
```
chungers: This has the same effect as the ADD removed (was in line 18) right?   Won't this copy the current directory (under docker/infrakit/*) into / and won't this cause problems with GOPATH since it's not laid out following the Go workspace directory conventions?

chungers:
@dgageot - currently Docker image builds happen only on master.   Let me test this out locally.

chungers:
I've patched in this PR locally and tested it...  Looks good.  Thanks for making this better!

I am adding a test script that will run the containers through the tutorial and verify the results.  It will be the integration test to run before publishing the images.

dockerfiles/Dockerfile.build
dgageot: + `COPY` is recommended over `ADD` (https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/)
+ Copying to `./` copies to the `WORKINGDIR`, so no issue, just some duplication removed.

dockerfiles/Dockerfile.build
dgageot: You can move it anywhere. Moving it before a `COPY` makes sure we minimize layer creation when a file changes in the repo

Issue #290, opened 2016-11-09
Support for different resource types


chungers:
There are more types of resources in a computing infrastructure landscape than just instances (compute).  The current focus on managing instances have led to the development of top level concepts like groups, instances, and flavors.  While these primitives and their operations (rolling update, scale, destroy) are quite general, we need to see how they can be applied to other resource types and if they don't fit, define new abstractions and models where they make sense.

The resource types that are important and need to be addressed include

  + networks
  + volumes
  + security groups / firewalls
  + load balancers

As state convergence of more resource types are addressed, it will also solve the *bootstrapping* problem.  Bootstrapping is defined as the process to set up the environment enough such that InfraKit can then take over and ensure the convergence of infrastructure state to user specification.  Carried to the extreme, the bootstrapping process can be reduced to the creation of a single node, the seed, from which everything else will be created, as InfraKit drives the entire infrastructure towards the user's specification.

chungers:
In terms of implementation, I see a few immediate tasks:

  + We need to define a schema for specifying a collection of resources.   This specification will also need to support dependencies amongst resources to be created.  This will be composed into a larger specification document with the existing specifications of groups.
  + Specify a new plugin type / API for creating different resources.   Do we want a single plugin that can handle different types of resources (e.g. subnets, security groups) or one for each type?  My sense is the former is simpler and better.  We can call it the *Resource plugin*.
  + Implement an engine that can perform dependency analysis and create a plan of actions such as creating resources via calls to the resource plugin.

Thoughts?  @nwt @fermayo

nwt:
> We need to define a schema for specifying a collection of resources. This specification will also need to support dependencies amongst resources to be created. This will be composed into a larger specification document with the existing specifications of groups.

For the collection of resources, I think we want something like
```
{
  "Resources": {
    "MyVpc": {
      "Plugin": "infrakit-resource-aws",
      "Type": "Vpc",
      "Properties": {
        "CreateVpcInput": {
          "CidrBlock": "10.0.0.0/16"
        }
      }
    },
    "MySecurityGroup": {
      "Plugin": "infrakit-resource-aws",
      "Type": "SecurityGroup",
      "Properties": {
        "CreateSecurityGroupInput": {
          "GroupName": "MySecurityGroup",
          "VpcId":  "{{.Resources.MyVpc.ID}}"
        }
      }
    },
}
```
where `{{.Resources.MyVpc.ID}}` is a dynamic value referring to the provider-assigned
[...]

PR #357, opened 2017-01-12
Bootstrapping and resources

nwt:
Here's a first stab at bootstrapping (#290) in the form of a resource plugin API and example along with a bootstrap plugin.

The bootstrap plugin implements the group API for expediency, and I've only implemented CommitGroup for now. I'm pretty sure we want a different API so we can give that method a better name (Bootstrap? CreateResources?). I'm pretty unsure what other methods will be useful.

A schema example is in docs/bootstrap.json. Make it go with
```sh
make build-binaries &&
mkdir -p resources && {
    build/infrakit-resource-file --dir resources &
    resource_pid=$!
    build/infrakit-bootstrap &
    bootstrap_pid=$!
} &&
build/infrakit group --name bootstrap commit docs/bootstrap.json &&
kill $resource_pid $bootstrap_pid
```


FrenchBen:
@nwt looks like you need to update the `vendor.conf` to include `github.com/twmb/algoimpl/go/graph`

chungers:
More comments to follow, but for starters, can you add some tests?

The code in `pkg/plugin/bootstrap`, that's all for the resource plugin, right?   Let's keep the naming consistent, so maybe rename the package to just `pkg/plugin/resource`?

The code that does the template and dependency analysis (via the topolgical sort) -- can we make that code reusable for any kind of templates that has the `{{.Some.Reference.To.Object.Property}}` notation?    Maybe add this functionality to the `pkg/template` where we generalize and standardized on using the golang template?   Just a thought.

docs/bootstrap.json
```diff
@@ -0,0 +1,20 @@
+{
+  "ID": "bootstrap",
+  "Properties": {
+    "Resources": {
+      "resource_a": {
+        "Plugin": "resource-file",
+        "Type": "a",
+        "Properties": {
+          "Note": "Depends on {{.Resources.resource_b.ID}}"
```
chungers: This reference here, `.Resource.resource_b.ID`, isn't explicit -- in the sense that I can't find anywhere in the document that says it satisifies the contract of having a field called `ID`.   This can be a problem in that it's impossible to validate this input (e.g, is it `ID` or maybe it's `Id`?)  

Another concern here is that this expression `.Resources.resource_b.ID` is closely coupled to the *structure* of this document and if there's a shuffling in schema all the expressions will have to updated to reflect that.  

Let me know if you agree.  I have some ideas how we can mitigate this.   
```

Comment (chungers) [inline docs/bootstrap.json]: Another thing I would like to discuss is whether it makes sense to use a Golang template here as an *input* to the plugins.   It's fine to use a Golang template and we plan to do more with it in the future, but I am less certain about plugins accepting a template (which may not be a valid JSON at all) as input because plugins communicate using JSON-RPC currently.

This example here is a valid JSON because your reference is inside a string, but if I want to reference a complex object then something like

```json
{
    "id" : {{.Resources.somethingElse.NumericValue}},
    "value" : {{.Resources.another_resource.ComplexObject}}
}
```
can't be marshaled and unmarshaled as proper JSON.

In general, golang templates work fine *around the edges* prior to becoming the input to the plugins.  I think the best thing here is to introduce a struct into the document like:

```json
{
    "id" : { "ref" : "somethingElse.NumericValue" },
    "value" : { "ref" : "another_resource.ComplexObject"}
}
```
This is similar to the approach taken by Cloudformation and Azure templates.  This is a valid JSON and is easy to parse (no regex in your bootstrapper code).    It can also be sent as-is as input via JSON-RPC.  When it is evaluated, the references can be substituted with actual value and rendered as proper JSON.

pkg/spi/resource/spi.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+// InterfaceSpec is the current name and version of the resource API.
+var InterfaceSpec = spi.InterfaceSpec{
+	Name:    "Resource",
+	Version: "0.1.0",
+}
+
+// Plugin is a vendor-agnostic API used to create and manage resources with an infrastructure provider.
+type Plugin interface {
```
chungers: This is virtually identical to the `instance` spi, except for the addition of a `resourceType` in some of methods.

That you had to duplicate code tells me there's an issue with our current model.  Perhaps we should visit the SPI of instance itself?   

What if we can do something to the instance spi itself or change the way a plugin exposes the RPC endpoints, etc. so that all the different resource types all implement the same spi -- compute, networks, volumes, etc., that would be nice.   Let me investigate this.  Thoughts?

docs/bootstrap.json
```diff
@@ -0,0 +1,20 @@
...
+        }
+      },
+      "resource_b": {
+        "Plugin": "resource-file",
+        "Type": "b",
```
FrenchBen: We may want to simplify this to follow what Microsoft does in Azure for their resource type:
`"type": "Microsoft.Storage/storageAccounts"`

The above would become:
```
"resource_b": {
  "Plugin": "resource-file/b",
  "Properties": {
  }
}
```

This would also us to re-use the instance type plugin

chungers: This is a pretty good idea.   

With a little of plumbing change, we can make it work so that a single plugin endpoint (eg. `aws`) can export multiple instance plugin objects, each one bound to a type (e.g. `subnet`).  So you could then use the Plugin name of `aws/subnet` and it will lookup the plugin with socket file `aws` and then route the RPC to the `instance.Plugin` implementation that's bound to the type name `subnet`.

There's something nice about this -- a single unified spi for immutable resource.  The existing instance plugins for provisioning compute instances is just a sub-type exported.   With the way we implement the RPC server and how the plugin objects are exposed, you can even introspect and see what 'types' an instance plugin can handle.

Let me put together a PR as a proof of concept.  If that looks good, then it will save us a lot of duplication of boilerplate code.

nwt: I like it, too.

Another option is to let plugins infer the type from what they find in Properties.

I don't know which I prefer, but both are better than a Type field.

chungers: See my PR #360 -- which implements this idea.

docs/bootstrap.json
nwt: I'd figured references would always occur inside strings, largely because in working with a prototype AWS resource plugin, IDs turned out to be the only useful dynamic value. That may well be an artifact of my use case, but it's worth thinking through now because IDs are the only dynamic value currently returned by Provision and DescribeInstances/DescribeResources.

nwt:
> looks like you need to update the vendor.conf to include github.com/twmb/algoimpl/go/graph

@FrenchBen: I sure did. Done in c88e0f186c64c5a588f9a319b8a6644482a5b3c2.

pkg/plugin/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,210 @@
...
+	"errors"
+	"fmt"
+	"regexp"
+	"sort"
+	"text/template"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/docker/infrakit/pkg/plugin/group/types"
+	"github.com/docker/infrakit/pkg/spi/group"
+	spi_resource "github.com/docker/infrakit/pkg/spi/resource"
```
nathanleclaire: FWIW, Go convention is to avoid `snake_case`, this would usually just be `spiresource` (standard lib has `httptest` for instance)

nwt: @chungers: Stick with snake_case or switch to notsoreadablecase?

pkg/spi/resource/spi.go
nathanleclaire: I'm just wandering in here but one idea I'd had about this kind of thing historically is to have a few different `interface`s which dictate the behavior of various resources.  For instance, the vast majority of instances can be `Create`-ed and `Destroy`-ed, but only some things (e.g., instances) can be `Stop`-ed and `Start`-ed...

Maybe your RPC server could accept attempted calls on a particular resource and throw issue if that particular plugin resource doesn't embed the given interface...

e.g.,

```go
// Create, Read, Delete
type CRDer interface {
    Create() error
    Read() (json.RawMessage, error)
    Delete() error
}

type StartStopper interface {
    Start() error
    Stop() error
}

type Instance interface {
    CRUDer
    StartStopper
}

type SecurityGroup interface {
    CRUDer
}
```

(It's worth considering more uncommon verbs like `Attach` as well... You can actually `Attach` and `Detach` security groups from an instance.  That's a totally different operation than just `Create`-ing some with given properties and could end up being pretty relevant for, e.g., VPC peering).

nwt:
> The code in pkg/plugin/bootstrap, that's all for the resource plugin, right? Let's keep the naming consistent, so maybe rename the package to just pkg/plugin/resource?

@chungers: Done in 0b338e95665d0ff97af87e539b979a95f494847f. I also renamed cmd/bootstrap to cmd/resource. Happy to undo that if that wasn't part of what you were thinking.

docs/bootstrap.json
nathanleclaire: FWIW, being able to interpolate strings directly (vs. `Ref` style) can be pretty nice, since you might want to do something like 

```json
{
    "Name": "{{.SwarmName}}-VPC-{{.Count}}"
}
``` 

to concat strings, whereas Azure ends up looking something like 

```json
{
    "Name": "[concat(variables('SwarmName'), 'VPC', count())]"
}
```

which I find a lot less readable.  The tricky part is to determine how nested you'll get, since [Terraform for instance](https://www.terraform.io/docs/configuration/interpolation.html) has somewhat a problem where these giant monstrosities of `join(split('/', concat(blah, blah, blah)))` get built up, and a nested way to do it might be nicer...  In that case, they partially have that issue because you need to use the string interpolation syntax to return other types such as lists.  I find their use of some things like `*` really hard to reason about as well...

Ultimately I think it all depends on use case.  Might be kinda nice to work towards getting a "short-hand" syntax for exclusively strings but also have `Ref` available when you need to do something slightly more complicated.

pkg/spi/resource/spi.go
chungers: @nathanleclaire  that's a good suggestion.  Currently, the instance has the notion of an attachment, but that attachment / detachment is implicit inside the implementation of the instance plugin.  I wonder if it would better to make that more explicit and use the interface mechanism you described here.

docs/bootstrap.json
nwt: My scope for the reference feature has been limited to resource IDs, which seems sufficient for bootstrapping a group. Does that scope need to expand?

chungers: Sorry for the late reply.  I think @nathanleclaire brought up some good points: having these structures at the start may be unnecessarily making things more complicated -- given that we are currently looking at just resource ID's.   Let's keep this as-is for now.  Thoughts?

nwt: Sounds good to me.

pkg/plugin/resource/resource.go
```diff
@@ -0,0 +1,200 @@
...
+
+	var sorted []string
+	for _, node := range g.TopologicalSort() {
+		sorted = append(sorted, (*node.Value).(string))
+	}
+
+	return sorted, nil
+}
+
+func (p *plugin) FreeGroup(id group.ID) error {
```
chungers: @nwt  - how about this...   let's not make the resource plugin implement the group SPI.   Rather, let's define a new SPI:
   1. Add a new package:  `pkg/spi/resource`
   2. Let's create a new `spi.go` and define something with what we for sure need for the resource plugin:
```go
type Plugin interface {
    Commit(spec *resource.Spec, pretend bool) (plan string, err error)
    Destroy(resource resource.ID) error
    DescribeResources() ([]resource.Description, error) 
}
```

In essence, the Resource plugin manages a collection of resources (which are created by the instance/<type> plugins).   It does things like analyze dependencies among this collection of resources in the spec JSON and calls the instance plugin for creation, etc.   It also provides the point of contact for end users to list, commit and destroy resources.

pkg/plugin/resource/resource.go
```diff
@@ -0,0 +1,200 @@
...
+	"errors"
+	"fmt"
+	"regexp"
+	"sort"
+	"text/template"
+
+	log "github.com/Sirupsen/logrus"
+	plugin_group "github.com/docker/infrakit/pkg/plugin/group"
+	"github.com/docker/infrakit/pkg/plugin/group/types"
+	"github.com/docker/infrakit/pkg/spi/group"
```
chungers: See comment below on creating a new SPI for resource.  Ideally this package won't have any import dependencies on the group pkg.   All the specs and types that currently reference the group pkg should just be recreated and put in the resource pkg.  This way, the resource plugin SPI is free to evolve on its own.

pkg/plugin/resource/resource.go
```diff
@@ -0,0 +1,200 @@
...
+type resource struct {
+	plugin instance.Plugin
+	config types.InstancePlugin
+}
+
+type plugin struct {
+	instancePlugins plugin_group.InstancePluginLookup
+}
+
+func (p *plugin) CommitGroup(config group.Spec, pretend bool) (string, error) {
```
chungers: Can you add a couple of tests for this with some example input?   In particular, the provisioning order (dependency) code can use a few tests to show usage and get some test coverage.  Thanks!

pkg/plugin/resource/resource.go
nwt: Sounds good to me.

pkg/plugin/resource/resource.go
nwt: Will do.

pkg/plugin/resource/resource.go
nwt: @chungers: Do we want to reuse `instance.ID` and `instance.Description` in this interface (instead of defining `resource.ID` and `resource.Description`)?

And do you envision that the resource plugin will maintain a mapping from `ID` to `instance.Plugin` for dispatching `Destroy`, similar to the group plugin?

chungers: Considering the resource plugin *delegates* resource CRD (create/read/destroy) to the instance plugin, using `instance.ID` may make sense.

On the issue of `Destroy` -- I am not so sure if destroying individual resources (the `Destroy` method with an id) makes a lot of sense now...  How about a `Destroy(spec *resource.Spec, pretend bool)` instead?   This is effectively the *inverse* of commit and we should be able to implement this by computing the *reverse* order of resource creation.    This should also work around the problem of maintaining association of Id to the instance plugin, right?

One thing we are ignoring for now is the issue that some resources cannot be destroyed if the code is running from *within* or *on* that resource -- eg. destroying a VPC in which the host that calls the `destroy` is running.   

What do you think?  Can you make this change and add the implementation to reverse the topological sort for destroying all the resources?

nwt: I like it. I'll get started on an implementation (with tests).

nwt:
@chungers: fbab57536aba096be631b32449146ece5876e3a9 implements the resource SPI proposed in https://github.com/docker/infrakit/pull/357#discussion_r96721556.

I still owe tests but wanted to share this sooner rather than later.

I also took a shortcut in cli/infrakit/resource.go by calling pkg/plugin/resource.NewResourcePlugin directly instead of using RPC. That's just temporary.

nwt:
@chungers: I've updated this thing to use pkg/template. Since that owns `{{...}}`, we need to pick something else for resource references.

Options I see:
1. `[[ .ResourceName ]]` (implemented in the update)
2. `[[ resource "ResourceName" ]]`
3. `{"resource": "ResourceName"}`

What are your thoughts?

nwt:
@chungers: Please take a look.

docs/resource.json
```diff
@@ -0,0 +1,20 @@
+{
+  "ID": "resource",
+  "Properties": {
+    "Resources": {
+      "resource_a": {
+        "Plugin": "instance-file",
+        "Type": "a",
+        "Properties": {
+          "Note": "Depends on [[ .resource_b ]]"
```
chungers: What if instead of using a different delimiter `[[`, we still use the same `{{}}` but introduce a new function instead?   

So we'd have something that looks like:

```
"Properties" : {
  "Note" : "Depends on {{ resource \"some/path/or/key\" }}",
}
```
With the `pkg/template` package it's easy to do, and you can bind different function implementations to the function name `resource` so that maybe in one stage of template processing, you are binding the keyword `resource` to a function that can construct a graph by adding a node whenever `resource` function is called.  And on another pass, you are returning actual, computed values from somewhere.

The escaping `"` is unfortunate, but if we switch to YAML in the future, it won't be a big eyesore.  I think this would at least make things look consistent with templates used elsewhere in the project.

Will this work?

nwt: @chungers: Works for me. It's implemented in 23503e2.

nwt:
@chungers:

1. Are you happy with what I've done in cmd/cli/resource.go?
2. If we're going to keep resource.Plugin.DescribeResources in the interface, I think we should change the signature. What would you say to
```DescribeResources(spec Spec) (plan string, err error)`?

**Comment by chungers** [comment]
@nwt 

> Are you happy with what I've done in cmd/cli/resource.go (calling pkg/plugin/resource.NewResourcePlugin directly), or would you prefer to have a cmd/resource/main.main?

Since this is a separate plugin /binary  / daemon, I think it's better that we follow the convention in the `cmd/manager` or `cmd/group`.   So I'd go with your second option -- `cmd/resource/main.go`, etc.

> If we're going to keep resource.Plugin.DescribeResources in the interface, I think we should change the signature. What would you say to this?
DescribeResources(spec Spec) (plan string, err error)

I think this is reasonable.  Let's go with your proposal.

**Comment by nwt** [comment]
@chungers: I just saw the comment in https://github.com/docker/infrakit/blob/master/pkg/manager/spec.go about avoiding maps in user-facing representations. In light of that, should we change the syntax from
```json
{
  "ID": "Fancy Resources",
  "Properties": {
    "Resources": {
      "A": {
        "Plugin": "instance-file",
        "Properties": {}
      },
      "B": {
        "Plugin": "instance-file",
        "Properties": {
          "Note": "Depends on {{ resource `A` }}"
        }
      }
    }
  }
}

```
to
```json
{
  "ID": "Fancy Resources",
  "Properties": {
    "Resources": [
      {
        "Name": "A",
        "Plugin": "instance-file",
        "Properties": {}
      },
      {
        "Name": "B",
        "Plugin": "instance-file",
        "Properties": {
          "Note": "Depends on {{ resource `A` }}"
        }
      }
    ]
  }
}
```
?

chungers:
@nwt  - Yes.  Sorry I missed this, but it is preferable that we use a list instead of a map.  This is because with a list we known every element struct contains all of the fields.  That the system may represent it internally as a map is really a matter indexing choices/ implementation detail.    So if you could follow this convention, that'd be great.

Let's use this convention
```
"Resources": [
      {
        "ID": "A",
        "Plugin": "instance-file",
        "Properties": {}
      },
      {
        "ID": "B",
        "Plugin": "instance-file",
        "Properties": {
          "Note": "Depends on {{ resource `A` }}"
        }
      }
  ]
``` 

So, for an infrakit object, we can always expect an `ID` field that is a peer of `Properties`, the same way we expect a `Plugin` field to denote the plugin responsible for its management.

nwt:
@chungers: I've switched from map to list as discussed in https://github.com/docker/infrakit/pull/357#issuecomment-287977799.

Is there anything else you'd like to see in here? Manager integration, maybe?

fermayo:
🎉 😍

Issue #398, opened 2017-02-14
BareMetal deployment


YujiOshima:
I'm eager Infrakit to support Bare metal deployment.
I created Canonical  MaaS plugin as PoC ( https://github.com/YujiOshima/infrakit-maas-plugin ) .
I am very happy if this is an official example, can I opne a PR? @chungers Thought?

chungers:
@YujiOshima  I think it would be very interesting to see an example of infrakit on bare-metal.  Go ahead a create a PR and add this to a directory like `examples/instance/maas` and we can discuss.    There are a few things to consider though... let me know if your thoughts:
  + Let's work together to make sure we have good documentation.
  + How do the users set up and play with this plugin?  It's for bare-metal so let's at least have some documentation / discussions on how to set things up?
  + Testing -- how do we test this plugin?

Another thing -- in the future we may move all the examples to repos under the 'infrakit' organization in GitHub.  This way we will be able to add more features to the individual plugin projects and evolve them.  To do this, however, require us to start thinking about strategies for automated testing / test harness / simulators, etc.  We'd also have to come up with ways to certify compatibility of plugins with the core infrakit controllers.   I'd love to hear what you think about these issues.

YujiOshima:
Thank you, we will open the PR soon.
* Documentation is an important issue. You should enrich not only the maas plugin but also the documents of various usage.
* In our assumed use case, we plan to deploy MAAS as a PXE server in the data center and deploy Infrakit and MAAS plugins at the same time. It is looking at the cooperation between on-premises and clouds that MAAS alone will not be completed. I expect Infrakit to construct an elastic infrastructure that links the on-premises and clouds together. (I'd like to create a multi-cloud group plugin for that!)
* Because of the nature of the bare metal, complete testing is difficult, so I think whether it is sufficient for testing the interface using go mock etc. first.

It is a good idea to separate the repository of plugins. In particular, it can be expected to improve the current situation that all dependent libraries of examples are also vendored.
For integration it is essential to run all plugin CIs with core infrakit controllers' commit as a hook. In our team dependency repository tests we create a new branch if the interface changes and we try to match the branch in the integration test (default is master). For example, if core infrakit create the branch called `apiv2` and commits it, CI will work on other plugins as a trigger, but if that `apiv2` branch is in that plugin it will be used. By doing so, you can also reduce the impact if you change the API.

chungers:
@YujiOshima  -- Thank you, all good points.

[...]

PR #411, opened 2017-02-23
Maas instance plugin

YujiOshima:
Support BareMetal deployment.
An Instance plugin for  [Canonical MaaS](https://maas.io/) node.
#398


chungers:
@YujiOshima  - nice work.  Just a general question...  does MAAS support the notion of detachable disk / volumes?   At least with the AWS plugin at https://github.com/docker/infrakit.aws it understands the `Attachment` of type `ebs`.   This ability is very interesting for cases of supporting stateful instances where attachable volumes store state and instances can come and go.  

Does this model make sense for MAAS or for bare-metal where disks/volumes can be attached /detached from compute instances?

YujiOshima:
@chungers Thank you!
That's good idea and important function for common "Pet" use case in the on-premises ! I didn't test that use case, but I think It can be done with iSCSI.
I will try to support volume store along with other functions. Now MaaS plugin is very incomplete, should be enhanced step by step.

Lennie:
In the meantime you could always use Docker volume plugin. For example just have 1 NFS share and create sub directories in your container when they don't exist yet. That way you don't need to request a share.

YujiOshima:
@Lennie Thank you for the comment. Only for docker swarm flavor and an app in the docker container, It is true. But for other flavor or outside of docker container (I don't know the specific use case, may be docker daemon config or something), you need Infrakit volume management.

Lennie:
Just saying workarounds are possible if you need to deploy soon and can't wait. Or need something for a dev. env. on a laptop. Also just realized you can create NFS server as a docker container as well.

But would be great to have a proper option for MAAS. But also don't forget the dev. laptop.

YujiOshima:
@Lennie Thank you for a good point. Yes, we need to do a lot of things before volume management.
At first I think that we can restrict the scope to only docker swarm or vanilla flavor.

examples/instance/maas/README.md
```diff
@@ -0,0 +1,68 @@
...
+Default login and password: `admin` `pass`
+
+Import boot images in `Images` tab.
+
+! Wait for import images. It takes some time. !
+
+Set up nodes.
+
+```
+./setup-nodes.sh
```
chungers: Is this file included in this PR?  I don't see it.   Would you mind including it?

examples/instance/maas/instance.go
```diff
@@ -0,0 +1,282 @@
...
+
+	for k, v := range labels {
+		tags[k] = v
+	}
+
+	encoded, err := types.AnyValue(tags)
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile(tagFile, encoded.Bytes(), 0666)
```
chungers: Files are written locally to store tags...  is there a way in MAAS to tag resources and query by tags?  https://docs.ubuntu.com/maas/2.1/en/intro-concepts#tags  

If we can reduce amount of state we need store locally, the better.     If there are ways to do tagging via the MAAS api, then we should consider making that change in a future PR?

examples/instance/maas/instance.go
```diff
@@ -0,0 +1,282 @@
...
+			if err := os.RemoveAll(machineDir); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// DescribeInstances returns descriptions of all instances matching all of the provided tags.
+func (m maasPlugin) DescribeInstances(tags map[string]string) ([]instance.Description, error) {
```
chungers: See comment above -- is there a way in MAAS to query by tags of the nodes?   Using files here is ok but we'd have to make sure these files are highly available.

examples/instance/maas/maas-vanilla.yml
```diff
@@ -0,0 +1,24 @@
+{
```
chungers: This file is named `.yml` but the content is a JSON...

examples/instance/maas/README.md
YujiOshima: I update the Readme. Now I have only a simple document, but I will enhance it with the addition of functions in the future.

examples/instance/maas/instance.go
YujiOshima: I think that is a better implementation. This time we implemented it in a simple way that does not depend on the api version, but we will change the information to keep it in a robust way without duplication.

examples/instance/maas/maas-vanilla.yml
YujiOshima: Oh..This is a mistake. Fixed.

examples/instance/maas/README.md
chungers: Sorry I was confused.  This is a file from your repo... I see now, so there's no need to include this.

chungers:
@YujiOshima - Before I can merge this PR, can you please add a few tests?    You can look at the `plugin_test.go` in the [terraform plugin](../tree/master/examples/instance/terraform/) as an example?   

Also, once this PR is accepted, would you mind open an issue on using MAAS tagging API to implement the instance state persistence?   I'd want to make sure that if we accept this PR this plugin will continue to be improved and is useful for those learning to work with InfraKit.  Thank you!

YujiOshima:
@chungers Yeah! I added a few tests!
> open an issue

Of course! I will do so soon. And I want to proceed while discussing.

YujiOshima:
I used a gomaasapi.testservice for test this plugin.
You can test using this for the basic function of MasS.

PR #422, opened 2017-03-03
Fix swarm flavor leak

### LINKED ISSUE #421: Swarm flavor plugin leaks connections
That is because we fail to close the docker client we use to connect.

dgageot:
docker client needs to be closed

Fixes #421


examples/flavor/swarm/flavor.go
```diff
@@ -228,6 +257,10 @@ func (s *baseFlavor) prepare(role string, flavorProperties *types.Any, instanceS
 			log.Warningln("Cannot prepare:", err)
 		}
 
+		if dockerClient != nil {
+			dockerClient.(io.Closer).Close()
```
YujiOshima: docker client ensures that transport.Client is closed before close ( https://github.com/docker/docker/blob/master/client/client.go#L183 ).
I am worried about io.Closer wouldn't care that. 
It is more safe to do same way as docker client.

dgageot: Hi @YujiOshima `io.Closer` is just an interface. the `Close` function that will be called is the one defined on `client.Client`. Anyways, I have to update this PR because I forgot things on the unit tests side.

PR #426, opened 2017-03-07
fix topic subscribing bug

### LINKED ISSUE #425: Event subscribing bug
With regard to event subscribe, topics that is not subscribed are also delivered from the server.
For example, `/ timer / sec / 10` and` / timer / sec / 100` are also subscribed if I subscribe `/ timer / sec / 1`.
This is because go-radix does a string-level search that does not treat `/` specially.

YujiOshima:
fix #425 

### DIFF
--- a/pkg/broker/server/sse.go
+++ b/pkg/broker/server/sse.go
@@ -160,9 +160,9 @@ func (b *Broker) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	}
 }
 
+//checkPath Compare the subscribed topic with the published topic to see if the message can be sent to the client.
 func (b *Broker) checkPath(subscribedPath string, publishedPath string) bool {
 	if subscribedPath != "/" {
-		// Make sure that the topic subscribed to by the client is the upper topic of the topic being notified or the topic exactly matched.
 		pPath := strings.Split(publishedPath, "/")
 		for i, sliceTopic := range strings.Split(subscribedPath, "/") {
 			if pPath[i] != sliceTopic {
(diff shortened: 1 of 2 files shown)


pkg/broker/server/sse.go
```diff
@@ -244,6 +245,15 @@ func (b *Broker) run() {
...
 					}
+					if key != "/" {
+						// Make sure that the topic subscribed to by the client is the upper topic of the topic being notified or the topic exactly matched.
+						notfyTopic := strings.Split(event.topic, "/")
+						for i, sliceTopic := range strings.Split(key, "/") {
+							if notfyTopic[i] != sliceTopic {
+								return false
+							}
+						}
+					}
```
chungers: Clever to do the check here.

Would you mind put this code in a function?   Something like  `if !checkPath(key, event.topic) { return false }`?

chungers: s/notfyTopic/notifyTopic/g

chungers:
@YujiOshima  Thank you for catching and fixing this bug and adding tests.

Once you move the check to a function I will merge it in.

I still think a better way is to intercept the subscribe call and reject a topic that doesn't conform to the spec of the Event plugin.  This way, we don't have to do this check at every outbound event publish.  It's also better UX since the client knows it tried to subscribe to an incorrect topic.

One place this might work, would be to add checks as an interceptor to the ServerHTTP handler.  This would be the place: https://github.com/docker/infrakit/blob/master/pkg/rpc/server/server.go#L132

Here we have references to all the `event.Plugin` implementations and we can wrap the `broker.ServeHTTP` handler with a handler that does proper topic validation before they are forwarded on to the real thing.  This would also give meaningful HTTP error code on subscription, rather than events quietly dropped to the floor.   Finally the `pkg/broker` doesn't have to have any knowledge of `/` convention for topics and can be used for something that uses a different topic naming convention (maybe `.object1.property1`).

pkg/broker/server/sse.go
YujiOshima: Thank you, moved to function!

PR #430, opened 2017-03-13
pkg/types/path Distinguish between /topicA  and /topicA/

YujiOshima:
@chungers In #429, behavior seems to be different from what we were discussing.
It can not distinguish between topic /A and topic /A/.
Therefore, subscribing `/toppicA` will always subscribe all of `/topicA/` and below subtopic, you can not subscribe only `/topicA`.
I added test `TestBrokerSubscriberExactMatchTopic`. This is the test I want to do in this PR.

### DIFF
--- a/pkg/broker/server/sse.go
+++ b/pkg/broker/server/sse.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -16,13 +17,19 @@ import (
 const patience time.Duration = time.Second * 1
 
 type subscription struct {
-	topic string
-	ch    chan []byte
+	topic      string
+	exactMatch bool
+	ch         chan []byte
 }
 
 type event struct {
 	topic string
 	data  []byte
+}
+
+type clientValue struct {
+	value      int
+	exactMatch bool
 }
 
 // Broker is the event message broker
@@ -80,6 +87,13 @@ func clean(topic string) string {
 	return topic
 }
 
+func checkExactMatch(topic string) bool {
+	if strings.LastIndex(topic, "/") == len(topic)-1 {
+		return false
+	}
+	return true
+}
+
 // Publish publishes a message at the topic
 func (b *Broker) Publish(topic string, data interface{}, optionalTimeout ...time.Duration) error {
 	any, err := types.AnyValue(data)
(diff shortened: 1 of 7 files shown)


pkg/broker/server/sse.go
```diff
@@ -239,13 +250,17 @@ func (b *Broker) run() {
...
+					chset, ok := value.(map[chan []byte]clientValue)
 					if !ok {
-
 						panic("assert-failed")
 					}
 
-					for ch := range chset {
+					for ch, value := range chset {
+						if value.exactMatch && event.topic != key {
+							fmt.Printf("skip notifier event.topic%v,  key %v\n", event.topic, key)
```
chungers: Can you remove this?

pkg/broker/server/sse.go
```diff
@@ -239,13 +250,17 @@ func (b *Broker) run() {
...
+							return false
+						}
+						fmt.Printf("notifier event.topic%v,  key %v\n", event.topic, key)
```
chungers: Same here.  Can you remove this line?

pkg/broker/server/sse.go
```diff
@@ -80,6 +87,13 @@ func clean(topic string) string {
 	return topic
 }
 
+func checkExactMatch(topic string) bool {
```
chungers: This can be written as one line right?

pkg/broker/server/sse.go
```diff
@@ -167,7 +181,7 @@ func (b *Broker) run() {
 			// Disconnect all clients
 			b.clients.Walk(
 				func(key string, value interface{}) bool {
-					chset, ok := value.(map[chan []byte]int)
+					chset, ok := value.(map[chan []byte]clientValue)
```
chungers: can't we just have a simpler value in the radix tree without the type `clientValue`?  

We really just need something like `map[chan[]byte]bool`, where the value bool is exactMatch or not.

pkg/broker/server/sse.go
```diff
@@ -16,15 +17,21 @@ import (
...
+	ch         chan []byte
 }
 
 type event struct {
 	topic string
 	data  []byte
 }
 
+type clientValue struct {
+	value      int
```
chungers: We probably don't need this struct type at all.  We only need to store the value in the radix tree as `map[chan []byte]bool` where bool is just the `exactMatch` value.  We don't need this `value` field...

pkg/broker/server/sse.go
YujiOshima: Sorry! I removed.

pkg/broker/server/sse.go
YujiOshima: Sorry! I removed.

pkg/broker/server/sse.go
YujiOshima: I wrote this because I was not sure how to use the original `int`.  Fixed, thanks!

Issue #414, opened 2017-02-28
Improve MaaS Plugin


YujiOshima:
The MaaS plugin needs to be implemented as a minimum function as a plug-in.
Refactor for robustness and current state support only vanilla flavor, so support swarm flavor first.
The functions to be planned are as follows.

- [ ] Refactor using MAAS tagging API instead of tag file.
- [ ] Support swarm flavor.
- [ ] Enhance deploy Documents.
- [ ] Automatically enlist new nodes.

We should support storage and network resources in the future.

YujiOshima:
@chungers Umm..I tried to make myself an Assignees, but I did not have the authority to set Assignees.

chungers:
@YujiOshima  -- Can you tell me what you mean by 'Support swarm flavor'?

This instance plugin has only two requirements to satisfy:  1. it should allow the instances to be tagged/ labeled, 2. compute instances should support userdata (e.g. cloudinit) or some scripts that gets executed when instances start up.

As long as an instance plugin can accept these (tags and userdata) and pass them to the platform for provisioning the instance, the Swarm Flavor should be reusable.

I hope this makes sense.  You shouldn't have to do anything extra -- or am I missing something?   Thanks!

YujiOshima:
@chungers Thanks! That is more accurate expression.
I was going to implement it that way, but as I was not convinced enough that alone, I wrote it as above (I think that support of userdata is sufficient!).

For the others, it is necessary to manipulate MaaS 's web interface only for deployment of MaaS server and enlist of new nodes.
I want to be able to operate as easily as possible. 
My Idea, deploying a MaaS server as a docker container.
If user already deployed a MaaS server, run only `build/infrakit-maas-plugin --apikey $APIKEY --url http://localhost:8080/MAAS `.
If user didn't have MaaS server, run `build/infrakit-maas-plugin --dockerhost /unix:///var/run/docker.sock `.
Then, MaaS plugin deploy MaaS server and connect to it.

YujiOshima:
I updated my comment.

chungers:
@YujiOshima  - if the user didn't have MaaS server running, what would the plugin do?   From your comment, it looks like you want it to start a Docker container running MaaS server?

Personally I think it's best to *not* have the plugin start that and just have the requirement that the MaaS server be running before the instance plugin starts up.  What do you think?   It's the more common use case anyway right?   People generally would have to have MaaS running somewhere as a service already...

YujiOshima:
[...]

PR #431, opened 2017-03-14
Update vendored gomaasapi

YujiOshima:
Related #414 
Current gomaasapi doesn't have testservice about tag operation.
Added in this commit ( https://github.com/juju/gomaasapi/pull/64 ).


YujiOshima:
@chungers Could you merge this? (Test looks failed at place unrelated with this PR. It may be unstable...)

PR #433, opened 2017-03-15
Adding configurable hostnames to the SoftLayer Terraform plugin

JacobFrericks:
Currently, the `softlayer_virtual_guest` `hostname` value is ignored and there is no support to define the hostname. This commit adds a new `@hostname_prefix` property; the unique infrakit timestamp is amended as a suffix. For example, if the `@hostname_prefix` is set to `given-hostname`, then the final value is `given-hostname-01234`. If the `@hostname_prefix` value is not set, then the current behavior is not altered.

### DIFF
--- a/examples/instance/terraform/plugin.go
+++ b/examples/instance/terraform/plugin.go
@@ -359,6 +359,23 @@ func (p *plugin) Provision(spec instance.Spec) (*instance.ID, error) {
 		}
 	}
 
+	// Use the given hostname value as a prefix if it is a non-empty string
+	if hostnamePrefix, is := properties.Value["@hostname_prefix"].(string); is {
+		hostnamePrefix = strings.Trim(hostnamePrefix, " ")
+		// Use the default behavior if hostnamePrefix was either not a string, or an empty string
+		if hostnamePrefix == "" {
+			properties.Value["hostname"] = name
+		} else {
+			// Remove "instance-" from "instance-XXXX", then append that string to the hostnamePrefix to create the new hostname
+			properties.Value["hostname"] = fmt.Sprintf("%s-%s", hostnamePrefix, strings.Replace(name, "instance-", "", -1))
+		}
+	} else {
+		properties.Value["hostname"] = name
+	}
+	// Delete hostnamePrefix so it will not be written in the *.tf.json file
+	delete(properties.Value, "@hostname_prefix")
+	log.Debugln("Adding hostname to properties: hostname=", properties.Value["hostname"])
+
 	switch properties.Type {
 	case "aws_instance", "azurerm_virtual_machine", "digitalocean_droplet", "google_compute_instance":
 		if t, exists := properties.Value["tags"]; !exists {
@@ -370,9 +387,6 @@ func (p *plugin) Provision(spec instance.Spec) (*instance.ID, error) {
 			}
 		}
 	case "softlayer_virtual_guest":
-		log.Debugln("softlayer_virtual_guest detected, adding hostname to properties: hostname=", name)
-		properties.Value["hostname"] = name
-
 		if _, has := properties.Value["tags"]; !has {
 			properties.Value["tags"] = []interface{}{}
 		}
--- a/examples/instance/terraform/plugin_test.go
+++ b/examples/instance/terraform/plugin_test.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"encoding/json"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"sort"
+	"strings"
 	"testing"
 
 	"github.com/docker/infrakit/pkg/spi/instance"
@@ -13,10 +15,12 @@ import (
 )
 
 func TestUsage(t *testing.T) {
+	// Test a softlayer_virtual_guest with an @hostname_prefix
 	run(t, "softlayer_virtual_guest", `
 {
         "type": "softlayer_virtual_guest",
         "value": {
+          "@hostname_prefix": "softlayer-hostname",
           "cores": 2,
           "memory": 2048,
           "tags": [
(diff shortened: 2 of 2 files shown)


JacobFrericks:
As far as code coverage, I noticed that the tests for this are checking to be sure there are no errors. Since this PR introduces some logic based on if a specific property value exists, would it be acceptable to add another SoftLayer test json for the `run` method? Or is that overkill just to test one property value?

YujiOshima:
I think that it is better to add it.
Because you can specify the state you expect by writing a test.
Please write down the purpose of json in the comment.

examples/instance/terraform/plugin_test.go
```diff
@@ -17,6 +17,7 @@ func TestUsage(t *testing.T) {
 {
         "type": "softlayer_virtual_guest",
         "value": {
+          "hostnamePrefix": "softlayer-hostname",
```
YujiOshima: This property name should be a snake case according to other properties.

JacobFrericks:
Thank you for your comment @YujiOshima! I'll make these changes asap!

JacobFrericks:
Looks like this failed on building vagrant (which is not touched in this PR)
```
?   	github.com/docker/infrakit/examples/instance/vagrant	[no test files]
time="2017-03-16T14:27:54Z" level=info msg="Connected: topic=/local/ => 1 registered clients, ch=0xc42001a6c0" 
time="2017-03-16T14:27:54Z" level=info msg="Connected: topic=/local/ => 2 registered clients, ch=0xc42001a9c0" 
time="2017-03-16T14:27:54Z" level=info msg="Connected: topic=/local/time/ => 3 registered clients, ch=0xc4200658c0" 
time="2017-03-16T14:27:56Z" level=info msg="Connected: topic=/local/ => 1 registered clients, ch=0xc4201583c0" 
time="2017-03-16T14:27:56Z" level=info msg="Connected: topic=/local/instance1 => 2 registered clients, ch=0xc42013d9e0" 
time="2017-03-16T14:27:56Z" level=info msg="Connected: topic=/local/instance => 1 registered clients, ch=0xc42017cb40" 
time="2017-03-16T14:27:56Z" level=info msg="Connected: topic=/local/instancetest => 2 registered clients, ch=0xc420192000" 
time="2017-03-16T14:27:56Z" level=info msg="Closing client connection at /local/ ch= 0xc4201583c0" 
time="2017-03-16T14:27:56Z" level=info msg="Closing client connection at /local/instance1 ch= 0xc42013d9e0" 
time="2017-03-16T14:27:56Z" level=info msg="Broker finished" 
time="2017-03-16T14:27:56Z" level=info msg="Connected: topic=/local => 1 registered clients, ch=0xc4201ca300" 
time="2017-03-16T14:27:56Z" level=info msg="Connected: topic=/local/instance => 2 registered clients, ch=0xc4201c8420" 
time="2017-03-16T14:27:56Z" level=info msg="Connected: topic=/local/ => 3 registered clients, ch=0xc4201c8540" 
time="2017-03-16T14:27:56Z" level=info msg="Closing client connection at /local/instance ch= 0xc42017cb40" 
time="2017-03-16T14:27:56Z" level=info msg="Closing client connection at /local/instancetest ch= 0xc420192000" 
time="2017-03-16T14:27:56Z" level=info msg="Broker finished" 
panic: no data: [100 97 116 97 58 32 10]
```

I ran `make ci` and it built fine locally. Is there any way to rebuild this? I get the message that says "You need write permission to trigger builds". Can I have someone do that for me?

kaufers:
Another implementation would be to simply use the current `hostname` attribute and amend a unique suffix (currently the `hostname` is totally ignored). I'm not sure if it's better to add a new `hostnamePrefix` attribute (and ignore `hostname`) or honor `hostname` and treat it as a unique prefix that infrakit adds a unique suffix to. Thoughts?

examples/instance/terraform/plugin.go
```diff
@@ -370,8 +370,25 @@ func (p *plugin) Provision(spec instance.Spec) (*instance.ID, error) {
 			}
 		}
 	case "softlayer_virtual_guest":
-		log.Debugln("softlayer_virtual_guest detected, adding hostname to properties: hostname=", name)
-		properties.Value["hostname"] = name
+		// Use the given hostname value as a prefix if it is a non-empty string
```
chungers: I think we can move this *entire* block to be outside the `switch` so that it can work for *all* resource types.   This is useful in general and the magic property is removed anyway.

Also, to indicate this `hostname_prefix` is a magic property, how about adopting a convention like `@hostname_prefix`?  This would make it easy for user to know which properties are essentially processing directives and not actual data.

chungers:
@JacobFrericks  @kaufers 

Nice work.  Thanks!

I think adding a `hostname_prefix` attribute gives less surprise to the user.   If the user has specified `hostname`, it will be surprising to see it mutated by the plugin into something else.  

That said, this new `hostname_prefix` attribute, which is interpreted specially only by a certain type of vm resource (`softlayer_virtual_host`), is a bit of magic that I think it's better to make it very obvious -- hence my suggestion to use something like `@hostname_prefix` to distinguish it from other attributes.  Also, I think it's something we can move outside the switch statement and make it work for all resource types.

Another idea is to turn all string values itself into templates so you'd have something like
```
hostname : myfancyprefix-{{ .ID }}-andmyfancysuffix
```
This will take more thought and we can implement later.

JacobFrericks:
@chungers do you foresee any complications using this with AWS, GCP, or Azure? I don't have much experience with those, to be honest.

examples/instance/terraform/plugin.go
```diff
@@ -370,8 +370,25 @@ func (p *plugin) Provision(spec instance.Spec) (*instance.ID, error) {
...
+		var hostnamePrefix string
+
+		switch properties.Value["hostname_prefix"].(type) {
```
chungers: probably just

```
var hostnamePrefix string
if v, is := properties.Value["@hostname_prefix"].(string) {
      properties.Value["hostname"] =  // do something here
}
// ...

delete(properties.Value, "@hostname_prefix")
...
```

chungers: Assuming we can use `@hostname_prefix` without problems, I think this should be fine.  All this extra processing is triggered by the presence of only the `@something` directive in the properties and only when a related `hostname` field is present...  So I think it's pretty safe.

JacobFrericks: Updated, can you take a look?

JacobFrericks:
Updated with the comments from @chungers and @YujiOshima

JacobFrericks:
@chungers can you take a look at the above message from @GordonTheTurtle? My commit is signed, so I'm not sure what I can do to fix this.

chungers:
@JacobFrericks  - if you've already signed it then it's fine.

examples/instance/terraform/plugin.go
```diff
@@ -359,6 +359,23 @@ func (p *plugin) Provision(spec instance.Spec) (*instance.ID, error) {
...
+			properties.Value["hostname"] = name
+		} else {
+			// Remove "instance-" from "instance-XXXX", then append that string to the hostnamePrefix to create the new hostname
+			properties.Value["hostname"] = fmt.Sprintf("%s-%s", hostnamePrefix, strings.Replace(name, "instance-", "", -1))
+		}
+	} else {
+		properties.Value["hostname"] = name
+	}
+	// Delete hostnamePrefix so it will not be written in the *.tf.json file
+	delete(properties.Value, "hostnamePrefix")
```
chungers: This should be `@hostname_prefix` right?

JacobFrericks: Good catch, yes it should be. I updated and rebased.

examples/instance/terraform/plugin_test.go
```diff
@@ -113,6 +172,22 @@ func run(t *testing.T, resourceType, properties string) {
...
+	var testingData interface{}
+	json.Unmarshal([]byte(properties), &testingData)
+	m := testingData.(map[string]interface{})
+	value, _ := m["value"].(map[string]interface{})
+
+	// If a hostname was specified, the expectation is that the hostname is appended with the timestamp from the ID
+	if value["@hostname_prefix"] != nil && strings.Trim(value["@hostname_prefix"].(string), " ") != "" {
+		newID := strings.Replace(string(*id), "instance-", "", -1)
+		expectedHostname := "softlayer-hostname-" + newID
+		require.Equal(t, expectedHostname, props["hostname"])
```
kaufers: Can we also assert that the `@hostname_prefix` is not part of the `props`?

JacobFrericks: Updated! See line 191.

PR #434, opened 2017-03-16
Use MaaS tag api for instance tag instead of file.

YujiOshima:
#414


chungers:
This is nice.  Thanks for doing this.

Is the MAAS api easy to mock?  If so some unit tests would be nice, since there's a bit of code in the processing of tags that's introduced in this PR.

YujiOshima:
Yes, I will add some more tests.

YujiOshima:
Updated! I added some tests.

Issue #83, opened 2016-06-27
Systematic logging support


chungers:
Right now logging is pretty ad-hoc.  Other than using logrus here and there, there is no pattern or systematic support for provisioner developers.  `docker-machine` exposes logger for driver developers to use and libmachete should as well.

linsun:
We really need to enhance the logging support.  First of all, all logging should have timestamp.  2nd, the current code has little or no log, it is really hard to debug things when stuff goes wrong.

wfarner:
@linsun completely agreed that logging is too sparse right now.  This issue relates to a logging framework for use in plugins, so i've filed #269 to track improvements to the thoroughness of logging.

chungers:
#440 implements this

PR #440, opened 2017-03-22
New logging infrastructure

chungers:
This PR starts the slow move to switch to a new logging package:
+  logrus is pretty tired and I miss filename and line numbers reporting in glog, but glog doesn't promote structured logging and isn't file rotation safe.  
+ use this https://godoc.org/gopkg.in/inconshreveable/log15.v2 instead.  This package is lightweight, supports composition, allows hierarchical loggers and can show call stack.
+ created a new `pkg/log` package to provide common defaults and configuration of the loggers.  

For normal development workflow, the `pkg/log` init() function will setup a JSON, multi-line, logger with callstack.  This is useful for logging in unit tests and line numbers are incredibly helpful.  

For running in CLI / production, a `pkg/log.(ProdDefaults)` provides a set of common defaults (e.g. Terminal color output, logfmt output, log levels.  These can be adjusted by additional CLI flags -- see `cmd/manager/main.go` for example.  Going forward,  we will slowly switch everything over to this new setup.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -1,15 +1,16 @@
 package main
 
 import (
-	goflag "flag"
+	"flag"
 	"os"
 	"path/filepath"
 	"time"
 
-	log "github.com/Sirupsen/logrus"
+	"github.com/Sirupsen/logrus"
 	"github.com/docker/infrakit/pkg/cli"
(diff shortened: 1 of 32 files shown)


chungers:
Closes #83

PR #441, opened 2017-03-22
push binaries to S3 bucket

chungers:


chungers:
Need to merge this because CircleCI settings won't work for PR branches.

PR #453, opened 2017-03-28
restrict hostname checks to softlayer only

chungers:
Apparently adding hostname prefix field in the properties for softlayer broke AWS because of extra `hostname` that is not allowed.  This PR moves the logic to a separate function and modified the tests to hopefully better isolate these special cases.

@JacobFrericks @ndegory

### DIFF
--- a/examples/instance/terraform/plugin.go
+++ b/examples/instance/terraform/plugin.go
@@ -332,6 +332,34 @@ func ensureUniqueFile(dir string) string {
 	return n
 }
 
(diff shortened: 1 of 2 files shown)


ndegory:
LGTM, with this patch I was able to successfully deploy on AWS using the Terraform plugin.

JacobFrericks:
LGTM I'm still able to spin up VSIs in SoftLayer using the magic key to have configurable hostnames.

chungers:
Thanks guys for confirmation!

Issue #449, opened 2017-03-27
Issues with tags on Softlayer


kaufers:
The characters that Softlayer supports for `tags` are limited: http://sldn.softlayer.com/reference/services/SoftLayer_Tag/setTags

> The characters permitted are a-z, 0-9, whitespace, _ (underscore), - (hypen), . (period), and : (colon). All other characters will be stripped away.

When the instance-plugin runs `terraform`, it detects a lot of deltas because the tags (which contains upper case, `/`, and `=`) cannot be set:

```# terraform apply
[WARN] /go/bin/terraform-provider-softlayer overrides an internal plugin for softlayer-provider.
  If you did not expect to see this message you will need to remove the old plugin.
  See https://www.terraform.io/docs/internals/internal-plugins.html
softlayer_virtual_guest.instance-1490622284: Refreshing state... (ID: 29990885)
softlayer_virtual_guest.instance-1490622284: Modifying... (ID: 29990885)
  tags.1516831048: "infrakit.group:workers" => "infrakit.group:workers"
  tags.2007221278: "name:instance-1490622284" => "Name:instance-1490622284"
  tags.2010887493: "" => "infrakit-link:f9Wb2FMIaD9bydoE"
  tags.2048522264: "" => "infrakit-link-context:swarm/cye5nlr6qgfjw33c80e05elnq/worker"
  tags.2127283095: "swarm-id:cye5nlr6qgfjw33c80e05elnq" => "swarm-id:cye5nlr6qgfjw33c80e05elnq"
  tags.2875799547: "infrakit-link-context:swarmcye5nlr6qgfjw33c80e05elnqworker" => ""
  tags.3780940857: "" => "infrakit.config_sha:aUAZntFm4TVOXdsJQUIAbhRezAQ="
  tags.583018450:  "infrakit.config_sha:auazntfm4tvoxdsjquiabhrezaq" => ""
  tags.89971309:   "infrakit-link:f9wb2fmiad9bydoe" => ""
softlayer_virtual_guest.instance-1490622284: Modifications complete (ID: 29990885)
```

kaufers:
Chatted with @chungers on this; the proposal is that the supported tag chars will be `[a-z][0-9]`, `:`, `-`, and `_`.

Currently, there is a tag like `Name:instance-1490622284` and this will change to:
```name:instance-1490622284
```

[...]

PR #461, opened 2017-04-03
Updating tags to only contain valid characters

JacobFrericks:
This PR changes the tag chars to be `[a-z][0-9], -, _` so that all tag values adhere to the supported tag chars defined in issue #449

### DIFF
--- a/examples/instance/terraform/plugin.go
+++ b/examples/instance/terraform/plugin.go
@@ -625,7 +625,9 @@ func terraformTags(v interface{}, key string) map[string]string {
 			value := fmt.Sprintf("%v", v)
 			if strings.Contains(value, ":") {
 				log.Debugln("terraformTags system tags detected v=", v)
-				vv := strings.Split(value, ":")
+				// This assumes that the first colon is separating the key and the value of the tag.
+				// This is done so that colons are valid characters in the value.
+				vv := strings.SplitN(value, ":", 2)
 				if len(vv) == 2 {
 					tags[vv[0]] = vv[1]
 				} else {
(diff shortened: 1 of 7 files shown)


examples/instance/terraform/plugin.go
```diff
@@ -382,8 +382,8 @@ func (p *plugin) Provision(spec instance.Spec) (*instance.ID, error) {
 	// set the tags.
 	// add a name
 	if spec.Tags != nil {
-		if _, has := spec.Tags["Name"]; !has {
-			spec.Tags["Name"] = string(id)
+		if _, has := spec.Tags["name"]; !has {
+			spec.Tags["name"] = string(id)
```
ndegory: On AWS (at least) the `Name` tag has a particular meaning, that's the first field displayed in the default header of the EC2 console page. Changing the case of the tag will result in an empty value in the Name column. So this change will have a huge impact on AWS deployments.

JacobFrericks: Thanks for your comment @ndegory! I've separated this out, so this shouldn't bother AWS. Please take a look at the updated code!

ndegory: thank! 👍

pkg/plugin/group/types/types.go
```diff
@@ -109,5 +110,10 @@ func (c Spec) InstanceHash() string {
 	hasher := sha1.New()
 	hasher.Write(stableFormat(c.Instance))
 	hasher.Write(stableFormat(c.Flavor))
-	return base64.URLEncoding.EncodeToString(hasher.Sum(nil))
+	encoded := base64.RawURLEncoding.EncodeToString(hasher.Sum(nil))
+	// Only valid characters are [a-z][0-9], _, - for support on specific platforms
+	encoded = strings.ToLower(encoded)
+	encoded = strings.Replace(encoded, "+", "-", 0)
```
chungers: I am not so sure if it's a good idea to do these string replacements here which destroys information. 

Instead, can we just switch to Base32 encoding (https://golang.org/pkg/encoding/base32/)?  Other than the case, it should have only the safe characters: http://www.garykessler.net/library/base64.html

PR #468, opened 2017-04-07
CLI extensions - playbooks

chungers:
New CLI extensions called 'Playbooks' allow user to package and define sets of templates and scripts and make the package available in the CLI as new subcommands and flags.  This complements the existing local extension mechanism of files/ directories by allowing 'playbooks' to be remote URL resources.  Currently the resource URL understand https/http/file resources but in the future we will add other protocols such as git or docker images.

A new command has been added: `infrakit playbook add | rm | ls`
This command is used to add new playbooks.  For example:

```shell
infrakit playbook add testdata file://$(pwd)/pkg/cli/remote/testdata/index.ikm
```
Adds the playbook called `testdata`:

```shell
infrakit -h

Available Commands:
.... # other entries 
 group         Access group plugin
  info          print plugin info
  instance      Access instance plugin
  manager       Access the manager
  metadata      Access metadata exposed by infrakit plugins
  playbook      Manage playbooks
  plugin        Manage plugins
  resource      Access resource plugin
  template      Render an infrakit template at given url.  If url is '-', read from stdin
  testdata      testdata  <=== added here
  util          Utilties
  version       Print build version information
```

Now remove the playbook:
```shell
infrakit playbook rm testdata
```


YujiOshima:
@chungers This is cool extension. I will try this! 
Since playbook is a new concept, We need Documents.
How about adding at least the description in this PR to cli's README as well?

chungers:
@YujiOshima  - thanks.  I am doing a little bit of code cleanup. Docs and examples coming soon!

PR #469, opened 2017-04-07
Docker instance plugin

ndegory:
Docker instance plugin for InfraKit

- creates containers based on specs built from Config, HostConfig and []NetworkResource.
- LogicalIDs are Docker hostnames instead of IP (Docker's DNS brings native resolution for the hostname)
- metadata from docker inspect

Can be used with any Docker images but can also be used with Docker in Docker containers in conjunction with the Swarm vanilla plugin.

### DIFF
--- a/examples/instance/docker/instance_test.go
+++ b/examples/instance/docker/instance_test.go
@@ -1,0 +1,116 @@
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/docker/docker/client"
+	"github.com/docker/infrakit/pkg/spi/instance"
+	testutil "github.com/docker/infrakit/pkg/testing"
+	"github.com/docker/infrakit/pkg/types"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	testNamespace = map[string]string{"cluster": "test", "type": "testing"}
+	tags          = map[string]string{"group": "workers"}
+)
+
+func TestInstanceLifecycle(t *testing.T) {
+	if testutil.SkipTests("docker") {
+		t.SkipNow()
+	}
+	defaultHeaders := map[string]string{"User-Agent": "InfraKit"}
+	cli, err := client.NewClient("unix:///var/run/docker.sock", "1.25", nil, defaultHeaders)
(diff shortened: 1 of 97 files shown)


YujiOshima:
@ndegory You need to edit vendor.conf and update vendored packages.
See https://github.com/rancher/trash

examples/instance/docker/instance_test.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+var (
+	testNamespace = map[string]string{"cluster": "test", "type": "testing"}
+	tags          = map[string]string{"group": "workers"}
+)
+
+func TestInstanceLifecycle(t *testing.T) {
+	defaultHeaders := map[string]string{"User-Agent": "InfraKit"}
+	cli, err := client.NewClient("unix:///var/run/docker.sock", "1.25", nil, defaultHeaders)
```
YujiOshima: May be you cannot access `docker.sock` in CircleCI test.

ndegory:
thanks for the review @YujiOshima , I'll take care about it.

examples/instance/docker/instance_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/docker/infrakit/pkg/types"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	testNamespace = map[string]string{"cluster": "test", "type": "testing"}
+	tags          = map[string]string{"group": "workers"}
+)
+
+func TestInstanceLifecycle(t *testing.T) {
```
chungers: CircleCI can't really run tests with docker reliably or at all.  Instead of removing this test, do this....

1. `import testutil "github.com/docker/infrakit/pkg/testing"`
2. Add a check here:
```
	if testutil.SkipTests("docker") {
		t.SkipNow()
	}
```
This will skip the test on CircleCI because there's an environment variable set when running tests on CircleCI (see the `circle.yml` file):
```
    SKIP_TESTS: "docker,etcd"
```

ndegory: thanks for the tip @chungers

PR #473, opened 2017-04-11
Common Spec schema and dependency ordering across all resource / group /plugin types

chungers:
## Background
This PR aims to establish a common data model for all the configurations for all controllers and plugins.   Previously we've had a general pattern of 
```json
{ 
  "ID" : "id",
  "Plugin" : "plugin-name",
  "Properties" : { }
}
```
for plugins and their properties.  However, this design has several drawbacks:
  + The schema doesn't describe the versioning of the plugin to use
  + The schema doesn't have a clear separation between user's specification and object state.
  + The schema doesn't have a standard way of capturing metadata like tags, names and ID's.  
  + The schema doesn't describe dependencies among objects.  This resulted in the Resource Plugin using a separate schema and `{{ resource }}` template expressions in field values.
  + The schema doesn't capture configurations parameters that are extrinsic to the managed objects but are nonetheless required for the *handlers* of these objects (the instance plugins, for example) to be configured properly.  Indeed these config parameters ended up being plugin CLI flags which are static and limiting.  For example, currently it's not possible to change the AWS instance plugins region or GCP plugins project without restarting the plugins.  These configuration parameter are hidden from the user and detracts from our goal of having a completely self-describing set of documents about the infrastructure, in every possible detail.

## Design
The basic design of the schema takes inspiration from the k8s api object schema, with adaptations that make sense for the way infrakit is architected (as a collection of active objects).  Special structures are introduced to allow user to express dependencies among managed objects.  Additional structures are added to support plugin configuration parameters that are outside of managed object properties

Here is an example YAML
```yaml
- class: aws/ec2-volume
   spiVersion: instance/v0.3.0
   metadata:
      name: disk1
      tags:
          project: mystack
   properties:
      size: 100
      encrypted: true
      iops: 200
   options:
       region: us-west-2
- class: aws/ec2-instance
   spiVersion: instance/v0.3.0
   metadata:
      name: host1
      tags:
          project: mystack
   template: https://infrakit.com/shared/aws-ec2-template
   properties:
          attachment :
             - id:  "{{ disk1.VolumeId }}"
                type: disk
          runInstancesInput:
             instanceType: t2micro
             keyName: myssh
   options:
       region: us-west-2
   depends:
        - class: aws/ec2-volume
           name: disk1
           selector: Status
           var: disk1         
```
In this example, there are two managed resources -- an ec2 instance `host1` and an ebs volume `disk1`: 
   + *class*: For specification of each object, we have the `class` field that indicates what kind of object/ handler to use -- `aws/ec2-instance` and `aws/ec2-volume`, respectively.  
   + *spiVersion*: specifies the version of the plugins that are responsible for managing the lifecycle of these objects.  
   + *metadata* - contains name, tags that are useful to describe the object.
   + *properties - captures the desired properties / configurations of the objects, as is currently the case.  
    + *template* - when `template` is present, the properties is first computed by rendering the template, followed by overrides in the `properties` section.  
    + *options* - The new `options` section allows configuration of the instance plugin (e.g. setting the `region`).  
    + *depends* - specifies the dependency of this object on others -- identified by the class and name.  A selector is used to apply a jmespath (http://jmespath.org/) and the result is bound the name *var* which can be referenced in the templates and properties.

## Implementation
1. In this PR I've added parsing and dependency calculation:  Given an arbitrary YAML / JSON document, extract all the objects that fit the above pattern.  Given a list of such specs, return a list of the specs in dependency order.

2. Refactor the Resource plugin to use this implementation

3. Migrate the `manager commit` to accept YAML documents that follow this schema.  
   + At this point there will be an UX change. 
   + Single schema for defining resources and groups.   


chungers:
For dependency specification, consider changing to

  + Use JSON Pointer ([IETF 6901](https://tools.ietf.org/html/rfc6901))  (golang implementation: https://github.com/xeipuuv/gojsonpointer) instead of JMESpath.
  + Support simpler binding specification by removing selector and just assume the entire Spec+Status (see the definition of golang `Object` struct) is bound to a var that is accessible inside the templates.  Like this:
```
   depends:
        class: Foo
        name: FooInstance
        var: /deps/provisionedFoo        
```
then inside the template, values about FooInstance can be retrieved via `{{ var "/deps/provisionedFoo/status/id" }}`

chungers:
This PR as it is contains tests that demonstrate the integration of FSM and dependency ordering of specs and the overlay of properties vs templates and results from dependency resolution.  There's still much more to apply this work for the generalized resource controller, but it is a good stopping point.

PR #477, opened 2017-04-11
[WIP] Add basic libvirt instance plugin

ijc25:
Supports basic VM creation but does not yet include storage or networking.
Hardcodes a bunch of things which should likely be exposed as properties, such
as the use of KVM and a serial console.

By default it targets `qemu:///session` which is a per-user session. To drive
the system instance run with `--uri qemu:///system`, this may require
authentication which is not yet supported (dependening on your desktop
environment it might result in a auth popup for each action).

The `Label` method is untested because I couldn't figure out how to cause it to
be called.

Signed-off-by: Ian Campbell <ian.campbell@docker.com>


ijc25:
Test failure is:
> command cd $WORKDIR && make ci took more than 10 minutes since last output

which I guess is unrelated.

YujiOshima:
@ijc25 Good Contribution! Please add some tests for libvirt instance!

rneugeba:
https://github.com/docker/infrakit.gcp has a bunch of tests with mocks and the like. Maybe an inspiration?

YujiOshima:
I think that terraform or maas is good in the sense that they have many tests.

>  I think we don't actually want to be starting real VMs in CI.

I agree.
If you can use libvirtd's mock, I think it is a good choice, but since CircleCI's tests are run on a container it may be out of privileges.
In that case, what about using go mock(https://github.com/golang/mock)? I think that you can test by creating mock of `libvirt / libvirt-go` with go mock.

YujiOshima:
Ah @rneugeba looks referred same. `infrakit.gcp` is using gomock.

chungers:
The build inside Docker container failed...  compiling libvirt:

```
go build -o build/infrakit-instance-libvirt -ldflags "-X github.com/docker/infrakit/pkg/cli.Version=6d612f0 -X github.com/docker/infrakit/pkg/cli.Revision=6d612f0fe4b82cdbf733d9a4da51b24e171f5cf8 -X github.com/docker/infrakit/pkg/util/docker.ClientVersion=1.24" github.com/docker/infrakit/cmd/instance/libvirt
# pkg-config --cflags libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt
Package libvirt was not found in the pkg-config search path.
Perhaps you should add the directory containing `libvirt.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libvirt', required by 'virtual:world', not found
Package 'libvirt', required by 'virtual:world', not found
Package 'libvirt', required by 'virtual:world', not found
Package 'libvirt', required by 'virtual:world', not found
Package 'libvirt', required by 'virtual:world', not found
Package 'libvirt', required by 'virtual:world', not found
Package 'libvirt', required by 'virtual:world', not found
Package 'libvirt', required by 'virtual:world', not found
Package 'libvirt', required by 'virtual:world', not found
Package 'libvirt', required by 'virtual:world', not found
Package 'libvirt', required by 'virtual:world', not found
```

ijc25:
@chungers I'll take a look, thanks (LOL @ `pkg-config --cflags libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt libvirt`, it really wants that package!)

@rneugeba / @YujiOshima it's not clear to me that mocking is all that useful here: the interesting thing to test is whether the XML given to the libvirt APIs (as a bare string) is well formed and semantically meaningful etc, which I don't think mock is capable of (I'd have to manually write the expected XML in the mock, but whose to say that matches what libvirt wants).

I'm going to try and use the [`test://` URI endpoint](https://libvirt.org/drvtest.html), perhaps with a locally spawned instance of the daemon, and test against that.

ijc25:
Fixed the container build and CI is nopw passing. I still want to add some tests. NB Friday and Monday are public holidays here and I'm on PTO on Thursday, so it might not be until next week.

ijc25:
Forced update:
- Added basic tests (basic lifecycle and labelling) using the libvirt test driver
- Ramdisk is now optional
- DomainType is now configurable (kvm vs ???)
- Ignore unknown domains which do not have the expected metadata fields

@chungers my `TestLabel` and `plugin.Label` functions were cowritten with my own (perhaps^Wprobably flawed) interpretation of what the `Label` method is supposed to do...

YujiOshima:
@ijc25 Great! But some comments.

1, Why did you put the libvirt plugin code under `cmd/instance` instead of putting it under` /examples` like other plugins?

2, The labels looks to be keeped only at memory. So when the plugin stoped, the labels will be lost.
Ideally, label information should be had at libvirt (like MaaS plugin) but if there is no such interface, it is better you should write on the files for persistence.
And TestLabel should check libvirt or files.

YujiOshima:
@ijc25 I am sorry, I was misunderstanding, I edited my comment.

pkg/plugin/instance/libvirt/instance_test.go
```diff
@@ -0,0 +1,139 @@
...
+	require.NoError(t, err)
+
+	require.True(t, domainExists(t, conn, domname), "domain was not found in libvirt domains")
+
+	plugin := NewLibvirtPlugin(libvirtDrvTest)
+	require.NotNil(t, plugin)
+
+	require.False(t, instanceExists(t, plugin, instance.ID(domname)), "domain was found in plugin instances before label")
+
+	labels := map[string]string{}
```
YujiOshima: Set not only emply but also some labels and check it.
Set some label when provisioning and add or edit labels with Label function.

ijc25:
@YujiOshima:
> Why did you put the libvirt plugin code under cmd/instance instead of putting it under/examples like other plugins?

@chungers asked me to do it this way.

pkg/plugin/instance/libvirt/instance_test.go
ijc25: >  Set some label when provisioning and add or edit labels with Label function.

I'll do that when someone points me to the documentation regarding the correct/expected semantics of the `Label` method. Otherwise I'm just guessing how it supposed to work and writing the answers to my guesses in both the implementation and the test.

chungers:
@ijc25 can this work with virtualbox on MacOSX?    I think we can merge this to get it in and iterate.

ijc25:
@chungers 

> can this work with virtualbox on MacOSX?

Does libvirt run on MacOSX and support VB there? If so then it ought to be possible to make it work. Note that libvirt doesn't abstract away all of the hypervisor specific bits and I've written/tested this with Qemu/KVM under Linux plus the test driver only so it is likely there will be a need to parameterise some more things to support other hypervisors and depending on the functionality available under those other hypervisors perhaps adding support for things like booting from ISO (e.g. if VB doesn't support direct kernel boot). Should all be pretty minor updates compared with the scaffolding added here though.

Looks like you have a policy of back merging `master` into the PR branches before merging, so I won't proactively rebase this PR now but if you would like me to do so just holler.

chungers:
@ijc25 - let's merge this and iterate.  Thanks for starting this plugin!

PR #497, opened 2017-04-28
libvirt: Support metadata using the LinuxKit metadata format.

ijc25:
«text from the final commit's commit message, previous two just update vendoring»

This uses the same mechanism/format as used by infrakit-instance-hyperkit, that is, it creates a ISO file containing the spec.Init in a file named `config` at the top level and attaches that as a CDROM to the device.
    
The iso image is placed in a configurable libvirt storage pool and we make an effort to remove it when destroying the instance.
    
Signed-off-by: Ian Campbell <ian.campbell@docker.com>


pkg/plugin/instance/libvirt/instance.go
```diff
@@ -31,9 +38,10 @@ type infrakitMetadata struct {
 	// <infrakit xmlns="https://github.com/docker/infrakit">...</infrakit>
 	// See https://github.com/golang/go/issues/13400,
 	// https://github.com/golang/go/issues/9519 and various linked issues.
-	XMLName   xml.Name              `xml:"https://github.com/docker/infrakit infrakit"`
-	LogicalID string                `xml:"logicalid"`
-	Tags      []infrakitMetadataTag `xml:"tag"`
+	XMLName          xml.Name                  `xml:"https://github.com/docker/infrakit infrakit"`
+	LogicalID        string                    `xml:"logicalid"`
+	Tags             []infrakitMetadataTag     `xml:"tag"`
+	MetadataDiskInfo *infrakitMetadataDiskInfo `xml:"metadata-disk",omitempty`
```
chungers: I think the syntax is `xml:"metadata-disk,omitempty"`  This was giving an go vet error.

ijc25:
I had a look at adding some instance metadata to improve the coverage but unfortunately the libvirt test/mock backend does not support `virStorageVolUpload`.

chungers:
@ijc25 I will merge this in when the tests pass.   At some point, we will need to do a pass on getting docs up to date.  Any notes you have a running / set up / examples will be very helpful and appreciated.

PR #520, opened 2017-05-04
Prepend instance id to resource instance names in terraform plugin.

craigyam:
When specifying a storage resource for the terraform instance
plugin, each instance must have a unique name.  Append the instance
id to the storage resource name before writing it out.

Signed-off-by: Craig Yamamoto <craigyam@us.ibm.com>

### DIFF
--- a/examples/instance/terraform/plugin.go
+++ b/examples/instance/terraform/plugin.go
@@ -353,7 +353,7 @@ func (p *plugin) Provision(spec instance.Spec) (*instance.ID, error) {
 		return nil, err
 	}
 
-	vmType, vmName, properties, err := FindVM(&tf)
+	vmType, _, properties, err := FindVM(&tf)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


examples/instance/terraform/plugin.go
```diff
@@ -343,6 +371,16 @@ func (p *plugin) Provision(spec instance.Spec) (*instance.ID, error) {
 
 	id := instance.ID(name)
 
+	storageType, storageName, storageProps, err := FindStorage(&tf)
```
chungers: I wonder if this could be handled similar to the `optionalProcessHostname()` function in line 401.  Say, let's have a `optionalProcessStorageName()` method that gets triggered when there's a `vmType` vm that matches `VMSoftLayer`.   This way, we can be very clear about how this logic applies only to Softlayer resources.   Right now,  anything that's in the `StorageTypes` slice will be applied this naming scheme, and someone could come along and add a storage type other than SoftLayer and be surprised by the mutation of the storage name...

Let me know if this makes sense.

craigyam: @chungers One question - wouldn't the desired behavior be to rename the resource (similar to how the virtual guest resource is renamed)?  Since each instance gets its own tf.json, terraform apply will fail when resource names are repeated (I'm assuming that behavior is not specific to SoftLayer resources, but I don't have experience with others).  Any thoughts?

chungers: You have a point.  I suppose the intention here is a bit different from naming the `hostname`, actually.  In the hostname case, it's actually an SL-defined attribute call hostname that needs to be modified, so the logic to base it off the VMType of softlayer_virtual_guest is reasonable. 

In this case, I think we are really saying this:
*If there is a vm*, then all the accompanying resources all need to have their terraform resource names qualified by the name of the instance.  Effectively these resources like storage,etc, are in the namespace of the vm of which they are peripherals / attachments.

If we go with this logic, then what you have is very true.  I would even go so far as to say let's generalize it even more...  If there is a VM instance, all the names of all the resources will be qualified by the generated vm name (the `infrakit-xxxxx`).  

What do you think?  This way you won't even have to have all the code specific to storage, storageType, etc.

chungers: Actually I'd go one further....

Effectively each invocation of the `Provision` method creates a set of resources belong to the same namespace -- the `infrakit-xxxx`.  So we can even go as far as not checking for presence of a vm instance at all... and just rename every resource by qualifying the name of the resource with the generated `infrakit-xxxx`. 

So a vm instance becomes `infrakit-xxxx-<the_original_node_name_in_the_config>`.    A storage instance becomes `infrakit-xxxx-<the_original_storage_name_in_the_config>`, etc. etc. 

This will ensure we always have unique names for any invocation of the plugin for any resources. wdyt?

craigyam: That makes sense to me.  I'll update this PR to add this support.

craigyam: @chungers I have updated the PR.  Thanks!

PR #543, opened 2017-05-12
Support int and bool types as template CLI vars

### LINKED ISSUE #542: Support int and bool types as template CLI vars
The `infrakit template` CLI parses all `--var` parameter values as strings, this prevents passing a variable like `--var /foo/boolean=true` as a boolean or even `--var /foo/int=1` as an int. The template functions like `not` need the values to be booleans instead of strings.

kaufers:
Instead of assuming that all CLI template `--var` values are String, this adds support to convert the String value to an int or boolean.

Closes #542

### DIFF
--- a/cmd/infrakit/base/template.go
+++ b/cmd/infrakit/base/template.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"strconv"
 	"strings"
 
 	"github.com/docker/infrakit/pkg/cli"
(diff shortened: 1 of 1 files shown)


kaufers:
@chungers initially I was using the `infrakit/pkg/cli/context.go parseBool` function that converted standard string like `true`, `T`, `yes`, `Y`, `0` to the `true` boolean. However, this seemed like it accepted too many values (since all ints were being converted to booleans and many strings like `yes` were as well); with that information I thought that simply using the golang reserved `true`/`false` for boolean would prevent unintentional string-to-boolean conversions.

cmd/infrakit/base/template.go
```diff
@@ -131,7 +132,18 @@ func TemplateProcessor(plugins func() discovery.Plugins) (*pflag.FlagSet, ToJSON
 				key := strings.TrimSpace(kv[0])
 				val := strings.TrimSpace(kv[1])
 				if key != "" && val != "" {
-					engine.Global(key, val)
+					// Attempt to convert to int and bool types so that template operations
+					// are not only against strings. Only honor "true"/"false" for bool so
+					// that a string like "no" will remain a string.
+					if intVal, err := strconv.Atoi(val); err == nil {
+						engine.Global(key, intVal)
+					} else if val == "true" {
```
chungers: I see your point about not using the `parseBool` function.  In this case, there's a `strconv.ParseBool` that should do the job.. https://golang.org/pkg/strconv/#ParseBool

kaufers: Yep, will update to use that so that the parsing is consistent. Thanks for the review!

PR #551, opened 2017-05-18
Kubernetes flavor

YujiOshima:
To support Kubernetes.
This flavor based on [kubeadm](https://kubernetes.io/docs/admin/kubeadm/).

Limitation (Next step)
* Currently you should log in to master node for check the cluster status. We should get cluster status with Health func.
* You can't create multi manager. It is limited with kubeadm.

Signed-off-by: YujiOshima <yuji.oshima0x3fd@gmail.com>


examples/flavor/kubernetes/README.md
```diff
@@ -0,0 +1,120 @@
+InfraKit Flavor Plugin - Swarm
```
chungers: Title...

examples/flavor/kubernetes/flavor.go
```diff
@@ -0,0 +1,348 @@
...
+			},
+		},
+		{
+			Name:        "NETWORK_ADDON",
+			Description: []string{"Returns the kube network addon"},
+			Func: func() (interface{}, error) {
+				var pfPath string
+				switch c.flavorSpec.KubeNWAddOn {
+				case "flannel":
+					pfPath = "https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml"
```
chungers: I wonder if we can move these hard coded strings to the flavor's `Properties` so that we can specify something like
```
    "KubeNWAddOn" : {
           "Name" : "calico",
           "Path" : "http://docs.projectcalico.org/v2.2/getting-started/kubernetes/installation/hosted/kubeadm/1.6/calico.yaml"
     }
```
This will allow us to avoid hard coding, and all the configurations can be inside the same JSON which makes it easy to see how system is configured.

examples/flavor/kubernetes/README.md
YujiOshima: OMG! 😱
I will fix immediately.

examples/flavor/kubernetes/flavor.go
YujiOshima: OK! That's sounds good!
I would like to make it a property of the [kubernetes addon](https://kubernetes.io/docs/concepts/cluster-administration/addons/) without limiting to network add-ons.
```
"KubeAddOn" : {
           "Name" : "calico",
	   "Type" : "network",
           "Path" : "http://docs.projectcalico.org/v2.2/getting-started/kubernetes/installation/hosted/kubeadm/1.6/calico.yaml"
     }
```
If you do not set network add-on only, cluster does not become ready status, so if you do not set network add-on, I will display a warning.

examples/flavor/kubernetes/flavor.go
```diff
@@ -0,0 +1,360 @@
...
+			Name:        "BIND_PORT",
+			Description: []string{"Returns the kubeadm JoinToken object"},
+			Func: func() (interface{}, error) {
+				return c.flavorSpec.KubeBindPort, nil
+			},
+		},
+		{
+			Name:        "NETWORK_ADDON",
+			Description: []string{"Returns the kube network addon"},
+			Func: func() (interface{}, error) {
```
chungers: How about making this a function that takes the addon type as a parameter?  Then we can name the function `ADDON` and its signature is `func(addonType string) (interface{}, error)` and we only need to search the list of addons and return the path of the first one with the matching type.

examples/flavor/kubernetes/flavor.go
```diff
@@ -0,0 +1,360 @@
...
+					if a.Type == "network" {
+						aPath = a.Path
+					}
+				}
+
+				return aPath, nil
+			},
+		},
+		{
+			Name:        "VISUALISE_ADDON",
```
chungers: See comment above...  if the function above takes a parameter then we won't need this function here.

chungers:
@YujiOshima  - sorry a small comment / change to consider before we merge.  Also can you sign the commits...

YujiOshima:
@chungers Thank you for the review! I updated it and cleaned up commits. I rebased with latest master. PTAL

PR #552, opened 2017-05-19
Libvirt updates

ijc25:
Two updates:

* Pass an entire libvirt domain configuration in `spec.Properties.Domain`, allowing for more flexibility.
* Expose any assigned IP addresses in the `instance.Description` if asked too (`properties == true`).


YujiOshima:
Hi @ijc25 !
Flexible configuration look so nice!
Some points.

* Do we need to set all `Domain` config? Do you have any default value? Previous code looks having default cpu, memory config.
* Though previous code checked property `Kernel` exist, this PR looks does not. Is it OK?
* Maybe it is better to use version of `libvirt-go` after c3209e4.
 Ref https://github.com/libvirt/libvirt-go/blob/0e52f11fc51cb385bdc638bd4a731e433e610f99/connect.go

ijc25:
> Do we need to set all Domain config? Do you have any default value? Previous code looks having default cpu, memory config.

The new code doesn't set any defaults, that's true. I could add them if you think it is useful, neither the hyperkit nor packet instance plugins seem to do that though.

> Though previous code checked property Kernel exist, this PR looks does not. Is it OK?

There are valid VM configurations which do not boot a kernel directly, for example ones which boot BIOS or UEFI firmware and a bootloader on the disk so there would be no kernel given in that case.

> Maybe it is better to use version of libvirt-go after c3209e4.

This PR switches to v3.3.0 which is the latest release, there has only been that one commit (c3209e4 which you referenced) since and it is to do with go 1.5 compat in the lxc backend so probably not too interesting to us.

ijc25:
CI failure is:
```
# github.com/docker/infrakit/vendor/github.com/libvirt/libvirt-go
vendor/github.com/libvirt/libvirt-go/lxc.go:63: cannot use cfdlist (type *C.int) as type unsafe.Pointer in argument to func literal
vendor/github.com/libvirt/libvirt-go/lxc.go:85: cannot use coldfdlist (type *C.int) as type unsafe.Pointer in argument to func literal
```
which might be down to the new commit in libvirt-go: https://github.com/libvirt/libvirt-go/commit/c3209e4ba8b8dda65c85ca0ac04302e55895caf7. I'll add a patch to update to that and see.

YujiOshima:
@ijc25 Thank you for the reply.
Since the presence or absence of a default value is not essential, I think that it will be fine as it is. 
I was wondering if there is an implied default value.
> Kernel 
I understood, thanks.

ijc25:
@YujiOshima 

> I was wondering if there is an implied default value.

It's not impossible that there is one at the libvirt level. I actually don't know. I think it probably makes sense to leave it to libvirt to have defaults, rather than add a second layer of potentially contradictory defaults on top.

chungers:
@ijc25 - there are errors with the tests
```
?   	github.com/docker/infrakit/pkg/plugin/instance/image/gcp	[no test files]
time="2017-05-25T03:30:14Z" level=error msg="Creating Domain: virError(Code=1, Domain=20, Message='internal error: missing domain type attribute')" instance=infrakit-7adef51f 
--- FAIL: TestBasicLifecycle (0.00s)
	Error Trace:	instance_test.go:78
	Error:		Received unexpected error:
			virError(Code=1, Domain=20, Message='internal error: missing domain type attribute')
			Creating Domain
			github.com/docker/infrakit/pkg/plugin/instance/libvirt.libvirtPlugin.Provision
				github.com/docker/infrakit/pkg/plugin/instance/libvirt/_test/_obj_test/instance.go:283
			github.com/docker/infrakit/pkg/plugin/instance/libvirt.(*libvirtPlugin).Provision
				<autogenerated>:2
			github.com/docker/infrakit/pkg/plugin/instance/libvirt.TestBasicLifecycle
				/home/ubuntu/.go_workspace/src/github.com/docker/infrakit/pkg/plugin/instance/libvirt/instance_test.go:77
			testing.tRunner
				/usr/local/go/src/testing/testing.go:657
			runtime.goexit
				/usr/local/go/src/runtime/asm_amd64.s:2197
		
```

Can you have a look?

ijc25:
Sure, I somehow missed that in the logs. In hindsight it is rather suspicious that `pkg/plugin/instance/libvirt/instance_test.go` wasn't touched by this PR.

ijc25:
@chungers fixed and force pushed. I added a test case for that specific failure as well as for another I found having fixed the first one.

ijc25:
Rebased, passed `properties=true` to `DescribeInstances` and added `TestInvalidProperties` to improve coverage.

ijc25:
@chungers tests have passed ;-)

PR #560, opened 2017-05-23
close stop chan in rollingupdate.go Stop func

### LINKED ISSUE #559: CI frequently fails.
CI often fails with the `pkg/plugin/group` package test timeout.
In rollingupdate.go, if waitUntilQuiesced returns before catching stopchan, it will block with [Stop function](https://github.com/docker/infrakit/blob/master/pkg/plugin/group/rollingupdate.go#L162).

YujiOshima:
fix #559 
Signed-off-by: YujiOshima <yuji.oshima0x3fd@gmail.com>


YujiOshima:
CI passed! 😆
I hope that it will stabilize.

PR #585, opened 2017-06-10
Migrate GCP plugin implementation to main repo

chungers:
This PR moves the code and history of the GCP (Google) instance plugin (etc.) into the main infrakit repo.  With this, we are also starting a `pkg/provider` as the way to organize different infrastructure providers.  In the future, we will migrate the provider-specific code in `pkg/plugin/instance` to follow this convention, because this new convention will scale better as each provider will contain implementations of multiple infrakit SPI, not just the interface SPI alone.

  + Reorganizes the `docker/infrakit.GCP` repo and moves the files to `pkg/provider/google`.
  + Contribution history is preserved
  + Updated MAINTAINERS to merge entries from source repo

@dgageot @FrenchBen


Makefile
```diff
@@ -177,12 +179,15 @@ terraform-linux:
 	@echo "+ $@"
 	wget -O tf.zip https://releases.hashicorp.com/terraform/0.9.3/terraform_0.9.3_linux_amd64.zip && unzip tf.zip && mv terraform ./build
 
+<<<<<<< HEAD
```
FrenchBen: Invalid git merge :P 
Needs to be cleaned out first.

PR #596, opened 2017-06-16
Terraform apply polling updates

### LINKED ISSUE #594: Updates to the terraform apply polling
Currently, once any `Provision` is executed on the terraform instance provider, a goroutine is created that loops and executes `terraform apply` at a random interval between 0 and 10 seconds.

The motivation for the random interval was to prevent multiple managers from colliding, but this will be fixed with #556 since only the leader will be polling.

There are a few enhancements we can make in this area:
1. Add CLI arg `--polling-interval` to configure the interval so that it's predictable
2. Sleep for a few seconds on the very first apply attempt (when the goroutine is created)
3. Attempt to run `terraform apply` on-demand from the provision instead of always waiting for the polling.

We want to do 2 because the common trigger for the first apply is from a group commit and most groups will have more then a single member. Currently, the group plugin issues the `Provision` on the first member and `terraform apply` is only executed against that member. Then, immediately after the first `Provision` the rest of the members are `Provision`'d but they aren't created until the first member completes. Since terraform can handle multiple resources in parallel, we want to try and provision the initial members in parallel.

We want to do 3 so that we can get updates as soon as possible; this becomes important for longer polling intervals.  This allows us to run `terraform apply` when we know that something has changed yet still poll at a predicable frequency.

kaufers:
- Adds a new `--poll-interval` that specifies the `terraform apply` frequency.
- Sleep for 5 seconds on the first apply, allowing the initial commit (which typically has more then a single member) time to get multiple `tf.json` files created
- Adds interrupt support to force an apply on-demand (vs. waiting for the next poll iteration)
- Stops polling goroutine if the manager plugin exists and it is not a leader; this checking can be disabled by specifying a new `--standalone` CLI arg

Closes #594

### DIFF
--- a/pkg/provider/terraform/instance/apply.go
+++ b/pkg/provider/terraform/instance/apply.go
@@ -5,8 +5,6 @@ import (
 	"time"
 
 	log "github.com/Sirupsen/logrus"
-	"github.com/docker/infrakit/pkg/discovery"
-	"github.com/docker/infrakit/pkg/discovery/local"
 	manager_discovery "github.com/docker/infrakit/pkg/manager/discovery"
 	"github.com/docker/infrakit/pkg/util/exec"
 )
@@ -36,20 +34,27 @@ func (p *plugin) terraformApply() error {
 	go func() {
 		initial := true
 		for {
-			if !p.continuePolling() {
-				p.applyLock.Lock()
-				p.applying = false
-				p.applyLock.Unlock()
-				return
-			}
-
-			attempted, err := p.doTerraformApply(initial)
-			initial = false
+			execute, err := p.continuePolling()
+			// Only poll is there was no error; if there was an error re-try next cycle
 			if err != nil {
-				log.Errorf("Executing 'terraform apply' failed: %v", err)
-			}
-			if !attempted {
-				log.Infof("Can't acquire apply lock, waiting %v seconds", p.pollInterval)
+				log.Errorf("Failed to determine if polling should continue: %v", err)
+			} else {
+				if !execute {
+					p.applyLock.Lock()
+					p.applying = false
+					p.applyLock.Unlock()
+					log.Infof("Exiting terraform apply gorouting")
+					return
+				}
+
+				attempted, err := p.doTerraformApply(initial)
+				initial = false
+				if err != nil {
+					log.Errorf("Executing 'terraform apply' failed: %v", err)
+				}
+				if !attempted {
+					log.Infof("Can't acquire apply lock, waiting %v seconds", p.pollInterval)
+				}
 			}
 
 			select {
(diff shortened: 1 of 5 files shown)


pkg/provider/terraform/instance/apply.go
```diff
@@ -18,30 +20,60 @@ func (p *plugin) terraformApply() error {
...
+	p.pollChannel = make(chan bool, 1)
 	go func() {
+		initial := true
 		for {
-			if err := p.lock.TryLock(); err == nil {
-				defer p.lock.Unlock()
-				doTerraformApply(p.Dir)
-			} else {
-				log.Debugln("Can't acquire lock, waiting")
+			attempted, err := p.doTerraformApply(initial)
```
chungers: I am not sure if you need to go through the trouble of adding channels to interrupt this loop just to prevent `doTerraformApply` from doing its thing if the plugin is not running on the leader node.   

There is a function that can locate the `Manager` object running on the same node.  You can simply call that to locate the Manager and then call `IsLeader()` and if it's not a leader, simply don't exec the `doTerraformApply`.  At the next time, this loop polls, same thing.  It's the same as having a separate polling goroutine but less complicated.

The helper is here:  https://github.com/docker/infrakit/blob/master/pkg/manager/discovery/discovery.go#L23

To call this, you do need a lookup function of the plugins... This is not currently plumbed through but the `infrakit` CLI offers an example.  I will add comments in your `main.go` so you know how to wire that.

pkg/provider/terraform/instance/main.go
```diff
@@ -35,11 +36,12 @@ func main() {
 	name := cmd.Flags().String("name", "instance-terraform", "Plugin name to advertise for discovery")
 	logLevel := cmd.Flags().Int("log", cli.DefaultLogLevel, "Logging level. 0 is least verbose. Max is 5")
 	dir := cmd.Flags().String("dir", getDir(), "Dir for storing plan files")
+	pollInterval := cmd.Flags().Duration("poll-interval", 30*time.Second, "Terraform polling interval")
 	cmd.Run = func(c *cobra.Command, args []string) {
 		mustHaveTerraform()
 
 		cli.SetLogLevel(*logLevel)
-		cli.RunPlugin(*name, instance_plugin.PluginServer(NewTerraformInstancePlugin(*dir)))
+		cli.RunPlugin(*name, instance_plugin.PluginServer(NewTerraformInstancePlugin(*dir, *pollInterval)))
```
chungers: Instead adding the `pollInterval` param to the function `NewTerraformInstancePlugin`, you could just add the function here that is required for plugin lookup:  https://github.com/docker/infrakit/blob/master/cmd/infrakit/main.go#L97

Also be sure to add a few lines of code to initialize the plugin discovery like so: https://github.com/docker/infrakit/blob/master/cmd/infrakit/main.go#L63

Note here there's no need to worry about "remote" plugin look up in this case.  Just call the local plugin lookup like so: https://github.com/docker/infrakit/blob/master/cmd/infrakit/main.go#L106

Once you have the `func() plugin.Plugins` plumbed into the Terraform plugin, you can pass that to the manager discovery function `pkg/manager/discovery/Locate()` to get a handle on the local manager running on the same node.  Once you have that you can call `IsLeader()` and run / not run `doTerraformApply` as appropriate.

pkg/provider/terraform/instance/apply.go
```diff
@@ -18,30 +23,105 @@ func (p *plugin) terraformApply() error {
...
+func (p *plugin) continuePolling() bool {
+	// Always continue if standalone
+	if p.standalone {
+		return true
+	}
+	if err := local.Setup(); err != nil {
+		log.Errorf("Failed to setup local discovery: %v", err)
+		return false
+	}
+	plugins, err := local.NewPluginDiscovery()
```
chungers: How about moving this initialization to the function where the plugin is created, in `NewTerraformPlugin`?   This way `func() discovery.Plugins` is always available.

pkg/provider/terraform/instance/apply.go
```diff
@@ -18,30 +23,105 @@ func (p *plugin) terraformApply() error {
...
+
+// continuePolling returns true if polling should continue; this happens if either
+// the plugin is configured to be standalone or if the associated manager plugin
+// is the current leader
+func (p *plugin) continuePolling() bool {
+	// Always continue if standalone
+	if p.standalone {
+		return true
+	}
+	if err := local.Setup(); err != nil {
```
chungers: This one too... probably the most important to have this in the `NewTerraformPlugin`.   We should try to error out as early as possible (at creation time of the plugin) when things like creating plugin directories, etc. fail.  This will simplify this code because we don't have to worry about conditions that cannot happen by the time this code executes.

kaufers: I've moved the local plugin `Setup` and `NewPluginDiscovery` to the constructor of the tf instance plugin (these will `panic` if there are errors). The tf instance plugin will have an additional `func() discovery.Plugins` struct element that will be used during the apply to retrieve the manager plugin.

Note that this will only happen if `--standalone` is **not** set. Also, if any errors are encountered during the polling (unable to find the manager plugin) the goroutine will not exit; it will simply try again on the next iteration.

PR #601, opened 2017-06-21
Ingress controller for Docker Swarm and loadbalancer SPI

chungers:
This PR starts the development of the ingress controller. An ingress controller manages the routes that bring traffic from Internet to the instances in a group.   The controller implementation here is a generic implementation ported from Docker for AWS and Azure.

  + The loadbalancer SPI defines a L4 loadbalancer that supports multiple routes and backend instances
  + The current PR does not address L7 loadbalancing and routing.  A new SPI will be created in the future.
  + Code for a base controller, in `pkg/controller/ingress/swarm` provides a controller implementation that continuously checks for Docker Swarm services and publishes these services to the external loadbalancer that implements the `pkg/spi/loadbalancer.L4` SPI.

The controller here was previously implemented to work with public cloud's auto-scaling groups.  So while publishing and unpublishing of routes (ports / listeners) based on Swarm services have been implemented, some additional features have yet to be implemented:
  
  +  The registration and de-registration of backend nodes to the loadbalancer.   Specifically the controller needs to interface with the Group controller to get a list of instances and correlate the instances with the Swarm nodes via the join tag.  Once a new instance / new node has been detected, it will call the SPI's `RegisterBackend` to add a new backend node.  Similarly, `DeregisterBackend` will be called to remove a node that is no longer a Swarm node.  The implementation of this could use polling.  Alternative, the event mechanism could be used to trigger a scan of known instances so that we can reduce qps on the infrastructure backends.
  + The controller and loadbalancer SPI still needs to implement RPC so that they are consistent with the group controller + instance plugin pattern.   
  + Packaging and command line (a `main.go`) to be added.

@kencochrane @kaufers @linsun


kencochrane:
thanks @chungers looks good to me

linsun:
@chungers thank you for the pull req and nice summary!  I have one q for you or @kencochrane:

>  Specifically the controller needs to interface with the Group controller to get a list of instances and correlate the instances with the Swarm nodes via the join tag.

Does the lb-controller really need to know the exact list of instances that made up a particular service or can the lb-controller rely on the routing mesh provided by swarm mode to route to a single endpoint (either node:{node-port} or {service-name}) where the routing mesh can decide what the list of instances is?

chungers:
@linsun 

You're right that technically we only need to route traffic to one node in the group and the routing mesh in Docker Swarm will ensure the actual node where the service container is running gets the traffic.  However, this is not efficient and we're susceptible to outages with this single node / few nodes go down or get replaced during a rolling update.

Another point is that not all systems that infrakit will support come with a routing mesh, so to be generic, we need to support the ability to add/remove backends for all the nodes in a group.  

Finally, the marginal effort of adding register/de-register of 1 vs. N nodes is trivial, so it's something we should just do it right.   Thoughts?

linsun:
@chungers thank you for your reply -- i agree with you that infrakit will support systems without routing mesh so no disagreement on providing the ability to add/remove backends for all instances in a service.  I was curious on what is the best practice when routing mesh is supported like swarm mode or k8s.   The reason I was asking is that if we could leverage routing mesh, then we don't need to constantly update the lb as the instances within the service got scaled out or in or even had ip changes.

chungers:
@linsun - I think the best practice is still having more than one instance in a group added as the backends of a loadbalancer.  That said, having a routing mesh like Swarm mode is nice because as long as we have a few nodes registered we can ensure there are multiple paths available to reach the service containers. 

It's the responsibility of the ingress controller to figure out what instances to add / remove as backends to a loadbalancer.  You only need to implement the Register/Deregister methods of the SPI.   With this, the bulk of the work is placed on the controller in pkg/controller/swarm.

linsun:
@chungers thanks, that sounds good.  It is good to hear that ingress controller will be in charge of adding/removing backends to a LB, which I assume it is listening to Docker events to figure that out for a given service in the case of swarm.

linsun:
@chungers @kencochrane do you happen to have a Dockerfile or instruction on how to run this lb-controller as a docker container?  We could use that before it is fully integrated to infrakit like you have described as https://github.com/docker/infrakit/pull/601#issue-237378572

sandys:
Will the Docker Swarm ingress implement proxy protocol ? Kubernetes and Openshift are already doing it - we have been discussing this in the moby repo. 
https://github.com/moby/moby/issues/25526#issuecomment-328253782

PR #606, opened 2017-06-23
VMware vCenter and vSphere plugin

thebsdbox:
This is the initial PR for the VMware vCenter/vSphere plugin for Docker Infrakit.

This plugin uses the VMware govmomi library to interact with the SOAP XML APIs that allow VM provisioning on both vCenter and the vSphere hypervisor directly.

Code around the `Label` function and around tag search in the `Describe` function needs completing.  Along with an update to the `Makefile`


chungers:
@thebsdbox 
Thanks for the PR.  Would you mind updating the `vendor.conf` for the govmomi package and run `trash`?  You can install `trash` (https://github.com/rancher/trash) and run from the project folder.  Then add the missing dependencies in `vendor/` and commit.   Thanks!

PR #610, opened 2017-06-26
Implementation of the loadbalancer L4 RPC interface

kaufers:

### DIFF
--- a/pkg/rpc/loadbalancer/client.go
+++ b/pkg/rpc/loadbalancer/client.go
@@ -1,0 +1,127 @@
+package loadbalancer
+
+import (
+	"time"
+
+	"github.com/docker/infrakit/pkg/plugin"
+	rpc_client "github.com/docker/infrakit/pkg/rpc/client"
+	"github.com/docker/infrakit/pkg/spi/loadbalancer"
+)
+
+// clientResult implements loadbalancer.Result
+type clientResult string
+
+func (f clientResult) String() string {
+	return string(f)
+}
+
+// NewClient returns a plugin interface implementation connected to a plugin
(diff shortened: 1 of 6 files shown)


pkg/testing/loadbalancer/l4.go
```diff
@@ -0,0 +1,66 @@
+package instance
```
kaufers: Wrong package, should be `loadbalancer.

pkg/rpc/loadbalancer/client.go
```diff
@@ -0,0 +1,127 @@
...
+
+// clientResult implements loadbalancer.Result
+type clientResult string
+
+func (f clientResult) String() string {
+	return string(f)
+}
+
+// NewClient returns a plugin interface implementation connected to a plugin
+func NewClient(name plugin.Name, socketPath string) (loadbalancer.L4, error) {
```
chungers: I think we can leave them as they are without the L4 qualification.  When we come up with L7 we can reuse the types when we can. If we need to qualify the types we will just create packages like pkg/spi/loadbalancer/l4 and pkg/spi/loadbalancer/l7.   So we don't need to add L4 in the names yet.

pkg/rpc/loadbalancer/types.go
```diff
@@ -0,0 +1,95 @@
+package loadbalancer
+
+import (
+	"time"
+
+	"github.com/docker/infrakit/pkg/spi/loadbalancer"
+)
+
+// NameRequest is the rpc wrapper for the Name request
+type NameRequest struct {
```
chungers: See my comment above. We can add packages when we need to. So what you have here is good!

PR #617, opened 2017-07-03
vSphere Fixes

thebsdbox:
- Fixed an issue with Provisioning timing out after a single second
- Added in the capability to find the guest ID and add it to tags for
describing
- Fixed a networking bug, where the wrong check resulted in no
networking being added
- Updated the label error message
- Fixed a bug where the tags map would report info from the first
instance
- Numerous error message fixes.


YujiOshima:
@thebsdbox I think the purpose of separating createNewVMInstance into goroutine is to avoid HTTP client timeout.
But It looks there are two problems below.
* When VM creation failed, Infrakit won't retry VM creation until next health check as infrakit has no chance to know the result of the VM creation.
* Infrakit will try check VM status without check the result of VM creation and may try to create new VM if the vSphere did not ready.

How about you set environment variable INFRAKIT_CLIENT_TIMEOUT longer?

PR #631, opened 2017-07-28
Import current instance into terraform on startup

### LINKED ISSUE #533: Import existing resource into terraform
When creating a swarm we have a "seed" VM that we also want to be the initial manager node; this VM is created outside of Infrakit so terraform has no knowledge of it. We need a mechanism to import an existing resource that the terraform Infrakit instance plugin did not create so that this initial VM can be managed with the other managers in the group.

kaufers:
Adds support for the following optional CLI params:

* `--import-group-spec-url`: The group specification URL that contains a nested instance specification; the `.tf.json` file for the imported resource contains the properties in the instance specification
* `--import-instance-id`: The ID of the instance to import
* `--import-group-id`: Optional group ID that the imported resource should be tagged with

If these 2 options are set then (on plugin startup):
1. Parse the group spec
2. Decode the instance definition from the group spec
3. Import the current VM into terraform
4. Parse the output of `terraform show` to get the current attributes of the current VM
5. Create a `.tf.json` file using the properties in the instance spec with the real attributes of the current VM; ensuring that the correct group tags (`infrakit.group=<grp-id>` and `infrakit.config_sha=bootstrap`) are set
6. `terraform apply`

Note that this only occurs of the current VM is not already in the terraform state file.

The plugin startup process is blocked until this has completed and will fail to start if there is an error.

Closes #533


pkg/provider/terraform/instance/plugin.go
```diff
@@ -50,7 +53,7 @@ type plugin struct {
 }
 
 // NewTerraformInstancePlugin returns an instance plugin backed by disk files.
-func NewTerraformInstancePlugin(dir string, pollInterval time.Duration, standalone bool) instance.Plugin {
+func NewTerraformInstancePlugin(dir string, pollInterval time.Duration, standalone bool, bootstrapGrpSpecStr, bootstrapInstID string) instance.Plugin {
```
chungers: I have some ideas on organizing this.   These two params are probably better grouped together -- let's have a struct called `bootstrap` that has the fields `InstanceID` and `GroupSpecURL`...   Then have the struct be a parameter in the function.  All the logic here can then be moved into a function that takes a `bootstrap` struct as arg or have a method (where a `*bootstrap` is the receiver).  This would make the "constructor" function a little easier to follow and makes it easy to isolate all the bootstrap functionality for tests.   We also wont end up one day with a function that has 100 parameters ;)

If the `bootstrap` struct is optional, I'd make the function arg be a pointer `*bootstrap` and then check for nil value.   

By the way, can you add in the comments for this PR an example of the group spec you've used?   I noticed only a `TFormat` (from the instance plugin's properties) is used even though it required the entire group spec...  I personally prefer to have the code in the import / bootstrap be very specific about what it needs and then push all the extra parsing code to the code outside of this function.  In other words, if the logic here requires only a TFormat object, then have the caller figure out how to feed it, rather than giving it some big encompassing struct and have this doing all the work of parsing it and extracting just a little piece.   I think making explicit requirements on the data the function needs to do the work would make it much easier to follow what's going on.

kaufers: @chungers The group spec needs to match https://github.com/docker/infrakit/blob/master/pkg/spi/group/spi.go#L31-L42

For example, see https://github.com/kaufers/infrakit/blob/3c2436308bcc7424030fc2be542bdb715c346bf7/pkg/provider/terraform/instance/plugin_test.go#L2521-L2532

As a concrete example, I've been using:

```
{
  "ID": "managers",
  "Properties": {
    "Allocation": {
      "LogicalIDs": [
        {{ range $i, $_ := loop (int (var "/cluster/swarm/manager/size")) }}
        {{ if ne $i 0 }},{{end}}{{ quote (cat "mgr" (add1 $i) | nospace) }}
        {{ end }}
      ]
    },
    "Instance": {
      {
        "Plugin": "instance-terraform",
        "Properties": {
          "resource": {
            "ibm_compute_vm_instance": {
              "host": {
    "Flavor": {
      "Plugin": "flavor-swarm/manager",
        "Properties": {
        "InitScriptTemplateURL": "{{ $managerInit }}",
        "DockerRestartCommand": "service docker restart",
        "SwarmJoinIP": "{{ var "/cluster/swarm/join/ip" }}",
        "Docker" : {
          "Host" : "unix:///var/run/docker.sock"
        }
      }
    }
  }
}
```

kaufers:
An example of the group spec that is passed is:

```
{
  "ID": "managers",
  "Properties": {
    "Allocation": {
      "LogicalIDs": [
        {{ range $i, $_ := loop (int (var "/cluster/swarm/manager/size")) }}
        {{ if ne $i 0 }},{{end}}{{ quote (cat "mgr" (add1 $i) | nospace) }}
        {{ end }}
      ]
    },
    "Instance": {
      {
        "Plugin": "instance-terraform",
        "Properties": {
          "resource": {
            "ibm_compute_vm_instance": {
              "host": {
                ...
              }
            }
          }
       }
    },
    "Flavor": {
      "Plugin": "flavor-swarm/manager",
        "Properties": {
        "InitScriptTemplateURL": "{{ $managerInit }}",
        "DockerRestartCommand": "service docker restart",
        "SwarmJoinIP": "{{ var "/cluster/swarm/join/ip" }}",
        "Docker" : {
          "Host" : "unix:///var/run/docker.sock"
        }
      }
    }
  }
}
```

pkg/provider/terraform/instance/plugin.go
kaufers: @chungers see latest commit, it introduces
```
type bootstrapOptions struct {
	InstanceID   *string
	GroupSpecURL *string
}
```

PR #639, opened 2017-08-02
Simplify running of infrakit plugins with "inproc" plugin starter

chungers:
This PR attempts to simplify running of many plugins in infrakit:

  + Implements a new `inproc` launcher for starting a plugin in-process.  This hooks into the current `infrakit plugin start` mechanism.
  + Start moving all the little binaries for all the servers (e.g. group, flavors, etc.) into the core `infrakit` binary.  Eventually we will eliminate all the other binaries and have `infrakit` be the single binary for everything
  + Run multiple plugins in the same process.  Initially this will still produce socket files on disk and the discovery mechanism remains the same (based on unix sockets).  However, this will simplify running infrakit because there is only one process or container to manage for everything.  The user can always run multiple processes but eventually the default will be to run multiple plugins inside a single process that can provide all the monitoring and remediation actions.  HA setup remains the same -- just that each node only needs to run a single infrakit process.

TODO
   + Migrate *all* server binary 'main' to a more modular system in `pkg/run`.  See `pkg/run/group` for an example.
   + Update documentation to use `infrakit plugin start group-stateless` instead of `build/infrakit-group-default`, for example.
   + Start up all plugins based on declarative specification in the new spec schema (see `pkg/types/Spec`).  
   + Implement a single `infrakit up <url>` where <url> points to the configuration yaml.  Infrakit will periodically check for changes in the content of <url> and performs reconciliation.  The user will not be expected to edit the file and call `infrakit manager commit` manually.  Instead, the user simply commits the config to her repository of choice (e.g. git) and infrakit will fetch the content periodically and reconcile.


thebsdbox:
Woooo

PR #642, opened 2017-08-06
Initial RemoteBoot support for InfraKit

thebsdbox:
- DCHP / TFTP / HTTP support (services can be used and configured as needed)
- Autogenerating of iPXE scripts for chain booting
- Hardcoded for linuxkit kernel/initrd (will be modifyable shortly)
- Logging uses two different methods, will need converging
- pulliPXE will auto download the PXE chain bootloader, next will add the capability to hardcode paths in the chain bootloader (for use in places where DHCP already exists)
- Testing needs adding
- Vendoring has been added


thebsdbox:
Docs added 😀

PR #650, opened 2017-08-09
Update Readme.md to reflect latest changes

dernotte:
It looks like the readme.md is no longer in sync with the API. I had to made some changes from the copy/paste from the Readme.md to make it works.


chungers:
Thank you so much for contributing to an area where we need help the most!

PR #690, opened 2017-09-18
Honor default `Options` in the `run/v0` package

kaufers:
When parsing the `Options` we need to start with the defined defaults so that the default values are honored if the user does not specify a given property.

Currently, the values are defaulted to the type defaults (for example, empty stings, 0 for ints, etc.).

### DIFF
--- a/pkg/run/v0/manager/manager.go
+++ b/pkg/run/v0/manager/manager.go
@@ -94,16 +94,16 @@ func defaultOptions() (options Options) {
 	switch options.Backend {
 	case "swarm":
 		options.Backend = "swarm"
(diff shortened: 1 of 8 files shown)


pkg/run/v0/manager/manager.go
```diff
@@ -94,16 +94,16 @@ func defaultOptions() (options Options) {
 	switch options.Backend {
 	case "swarm":
 		options.Backend = "swarm"
-		options.Settings = DefaultBackendSwarmOptions
+		options.Settings = types.AnyValueMust(DefaultBackendSwarmOptions)
```
chungers: This is much better than before.  Thanks.

PR #691, opened 2017-09-18
Added support for getting CFN output values in aws metadata plugin

kencochrane:
This allows you to gather values from the cloudformation outputs for a stack. This is helpful when trying to get results from a stack and using it as inputs down stream.

in a template you could do something like this.

```
{{ var "/my/cfn/output" (metadata "var/export/cfn/config/Outputs/MyStackOutput/OutputValue" )}}
```

Signed-off-by: Ken Cochrane <kencochrane@gmail.com>

### DIFF
--- a/docs/controller/ingress/example.yml
+++ b/docs/controller/ingress/example.yml
@@ -26,7 +26,7 @@
 #m4fghruao79i        t2                  replicated          1/1                 nginx:latest        *:7777->80/tcp
 #
 # Remove the service
-# docker service rm test1
+# docker service rm t2
 #
 # infrakit simulator/lb1 routes ls  # should be empty
 #
--- a/pkg/provider/aws/plugin/metadata/funcs.go
+++ b/pkg/provider/aws/plugin/metadata/funcs.go
@@ -135,9 +135,19 @@ func cfn(clients AWSClients, name string) (interface{}, error) {
 		parameters[*p.ParameterKey] = p
 	}
 
+	// index outputs by name
+	outputs := map[string]interface{}{}
+	for _, p := range output.Stacks[0].Outputs {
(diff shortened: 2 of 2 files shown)


docs/controller/ingress/example.yml
```diff
@@ -26,7 +26,7 @@
 #m4fghruao79i        t2                  replicated          1/1                 nginx:latest        *:7777->80/tcp
 #
 # Remove the service
-# docker service rm test1
+# docker service rm t2
```
kencochrane: fixed typo

pkg/provider/aws/plugin/metadata/funcs.go
```diff
@@ -135,9 +135,19 @@ func cfn(clients AWSClients, name string) (interface{}, error) {
...
 
+	// index outputs by name
+	outputs := map[string]interface{}{}
+	for _, p := range output.Stacks[0].Outputs {
+		if p.OutputKey == nil {
+			continue
+		}
+		outputs[*p.OutputKey] = p
+	}
+
```
kencochrane: I tried to test locally, but not sure how to test, any suggestions would be helpful.

kencochrane:
It looks like the test failure isn't related to my change.

pkg/provider/aws/plugin/metadata/funcs.go
chungers: The only way I can see is to mock the API call but tbh it really isn't testing much given how simple the code fragment is.

kencochrane: Thanks

kencochrane:
I'm not sure why the test keep failing. It doesn't look like it is related to my change. I could be wrong.

PR #694, opened 2017-09-20
Add support for Option `Envs` parser and use in tf plugin

kaufers:
Adds support to pass custom env vars to the terraform plugin. The idea is these env vars will be the cloud provider credentials that should be included when running the `terraform ....` execs; this can be used to parse a `docker secret` file that contains the specific cloud provider credentials.

For example:
```
"Options": {
  "Envs": [
    "USER={{ $j := include `file:///run/secrets/softlayer` | jsonDecode }}{{ $j.username }}",
    "PASS={{ $j := include `file:///run/secrets/softlayer` | jsonDecode }}{{ $j.password }}"
  ]
}
```

The terraform instance plugin is also updated to use this Option


pkg/provider/terraform/instance/apply.go
```diff
@@ -249,9 +255,12 @@ func (p *plugin) handleFiles(fns tfFuncs) error {
 }
 
 // doTerraformStateList shells out to run `terraform state list` and parses the result
-func doTerraformStateList(dir string) (map[TResourceType]map[TResourceName]struct{}, error) {
+func (p *plugin) doTerraformStateList(dir string) (map[TResourceType]map[TResourceName]struct{}, error) {
```
craigyam: Since you added the receiver here, do you still need to pass in dir?  Can you use p.Dir instead?

craigyam: Same comment for doTerraformShow and doTerraformShowForInstance

kaufers: Thank @craigyam, good catch.

pkg/util/options/parser.go
```diff
@@ -0,0 +1,71 @@
+package options
```
chungers: Do you mind if we put this as a package in `pkg/provider/terraform`?    The options parsing here is pretty specific to terraform at the moment so it may make sense to have this code as a package within terraform.

kaufers: Sine we are using the native templating from the plugin config file this entire package is unnecessary; I will remove.

pkg/util/options/parser.go
```diff
@@ -0,0 +1,71 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"regexp"
+	"strings"
+
+	"github.com/docker/infrakit/pkg/types"
+)
+
+var fileRegex = regexp.MustCompile("^file://(/.*):(.*)$")
```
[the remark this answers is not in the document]
kaufers: Removing.

thebsdbox:
This is looking OK to me with the recent changes, so happy to merge.

PR #704, opened 2017-09-28
New ibmcloud type and volume auth instance plugin

kaufers:
Adds a new `ibmcloud` type with a single instance plugin that handles VM authorization to a storage volume. The instance plugin authorizies a VM ID to a volume (`Provision`), removes the VM authorization from a volume (`Destroy`), and returns the IDs of all VMs that are currently authorized to a volume (`DescribeInstances`).

This new plugin should be used with the enrollment controller to ensure that members of one or more groups are authorized to the volume.


pkg/provider/ibmcloud/plugin/instance/softlayerclient.go
```diff
@@ -0,0 +1,47 @@
+package instance
```
kaufers: @chungers Thoughts on the correct package for this?

PR #705, opened 2017-09-29
Added an AWS template function that returns AZs

kencochrane:
This PR adds an AWS template function that will return a list of availabilityZones in the current region for the given user.

This will be helpful when spreading out the ec2 instances across the different AZs in a region.

Signed-off-by: Ken Cochrane <kencochrane@gmail.com>


kencochrane:
Not sure why the tests failed, don't think it is related to my PR

PR #706, opened 2017-09-29
Fix CI issue with returned type

FrenchBen:
`nodes` is outputting a `%s` when the `NodeList` returns `[]swarm.Node`
Changed the output type to be `%v` to get CI to pass


kencochrane:
LGTM

PR #711, opened 2017-10-02
Added Ben to maintainers list

FrenchBen:


MAINTAINERS
```diff
@@ -36,6 +37,11 @@
 	Email = "david.gageot@docker.com"
 	GitHub = "dgageot"
 
+	[people.FrenchBen]
+	Name = "French Ben"
```
thaJeztah: Looks like you swapped your "first" and "last" name here, compared to the other repo 😉  https://github.com/docker/kitematic/blob/master/MAINTAINERS#L34

FrenchBen: haha which one is better? 🤔

thaJeztah: You're real name :innocent:

thaJeztah: *your* - damn you autocorrect

PR #714, opened 2017-10-06
Kube support + bootstrapping

chungers:
This PR:

  + Improved Kube support by allowing specification of the control plane nodes for the kube cluster via new `ControlPlane` field.  This is necessary for self-starting a kube cluster where all the nodes are part of the cluster (instead of the current N+1 configuration, where the 0th node is used to run just infrakit).  In this case, we allow specification of a set of nodes for a control plane (for HA infrakit) and the Kube control plane is a strict subset of those nodes (which are designated via the `ControlPlane` field).  If you want to use a single master kube, then specify the `ControlPlane` field with the logical ID (usually the IP address) of the kube master.  We will support multi-master HA set up for kube where each Infrakit node is also a kube master.  This PR support a single master.
  + Added a `infrakit util init` command.  This command takes a groups.json and renders the Init for a member of the group of your choice.  With this, you can get a Init script for the initial boot node.  With this, you can get the join tokens via the `kubeadm` inside the kubernetes flavor.  

Example:

On AWS if you have a single node.. you can then call `infrakit util init` as your cloud init or pipe to a shell:
```json
 "UserData" : {
  "Fn::Base64" : {
     "Fn::Join" : [ "", [
       "#!/bin/sh \n",
       {  "Fn::FindInMap": [ "AMISelector", { "Ref" : "PreinstalledDockerAMI" }, "userdata" ]  },
       "mkdir -p /infrakit\n",
       "docker swarm init\n",
       "docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v /infrakit:/infrakit ",
	"-e INFRAKIT_FLAVOR_KUBERNETES_CONFIG_DIR=/infrakit ",
	{"Ref":"Infrakit"},
        " infrakit util init --group-id kube-managers --start combo --start swarm --start kubernetes ",
            " --var /cluster/name=", {"Ref":"AWS::StackName"},
            " --var /cluster/size=", {"Ref":"ClusterSize"},
            " --var /infrakit/config/root=", {"Ref":"InfrakitConfigRoot"},
            " --var /infrakit/metadata/configURL=", {"Ref":"MetadataExportTemplate"},
            " --var /provider/image/hasDocker=", {
                      "Fn::FindInMap": ["AMISelector", { "Ref":"PreinstalledDockerAMI" }, "hasDocker"]
                   },
            " --var /infrakit/docker/image=", {"Ref":"Infrakit"}, " ",  {"Ref":"InfrakitConfigRoot"}, "/groups.json",    " | tee /var/lib/infrakit.boot | sh \n"
          ]]
         }
        },
```
In this example, in CloudFormation, the first node's init is generated by infrakit itself.  This ensures the first node has the same configurations as the second, and third nodes later.

The config looks like below.  Note the configuration of the manager is a combination of swarm mode and kubernetes, with kube control plane being a subset of the the swarm mode control plane / managers:
```
{
   "Plugin": "group",
    "Properties": {
       "ID": "kube-managers",
       "Properties": {
           "Allocation": {
               "LogicalIDs": {{ $managerIPs | jsonEncode }}
            },
           "Flavor": {
	       "Plugin" : "combo",
	       "Properties" : [
		    {
		       "Plugin": "swarm/manager",
		       "Properties": {
				"Attachments" : {
				    "{{index $managerIPs 0}}" : [
					{ "ID":"{{index $managerIPs 0}}", "Type":"ebs" }
				    ],
				    "{{index $managerIPs 1}}" : [
					{ "ID":"{{index $managerIPs 1}}", "Type":"ebs" }
				    ],
				    "{{index $managerIPs 2}}" : [
					{ "ID":"{{index $managerIPs 2}}", "Type":"ebs" }
				    ]
				},
				"InitScriptTemplateURL": "{{ $managerInit }}",
				"SwarmJoinIP": "{{ $swarmLeaderIP }}",
				"Docker" : {
				    "Host" : "{{ $dockerEngine }}"
				}
			    }
	            },
	            {
		        "Plugin": "kubernetes/manager",
			"Properties": {
			  "InitScriptTemplateURL": "{{ var `/infrakit/config/root` }}/kube-managers-init.sh",
			  "KubeAddOns": [
				    {
					"Name": "calico",
					"Path": "http://docs.projectcalico.org/v2.2/getting-started/kubernetes/installation/hosted/kubeadm/1.6/calico.yaml",
					"Type": "network"
				    },
				    {
					"Name": "kubedash",
					"Path": "https://raw.githubusercontent.com/kubernetes/dashboard/master/src/deploy/recommended/kubernetes-dashboard.yaml",
					"Type": "visualise"
				    }
				],
				"KubeBindPort": 6443,
				"KubeJoinIP": "{{ var `/cluster/kube/join/ip` }}",
				"KubeClusterID" : "{{ var `/cluster/name`}}",
				"SkipManagerValidation" : true,
				"ControlPlane" : [ "{{ index $managerIPs 0 }}" ]
			    }
			}
		    ]
                },
                "Instance": {{ $myGroups.Managers | jsonEncode }}
            }
        }
```
For an example of a self-bootstrapping Kube cluster... see [here](https://github.com/chungers/examples/tree/master/latest/kube)

TODO 
   + Implement node Taint / Drain for node terminations.  Apparently kube likes to keep the nodes hanging around even after they have been terminated.  Swarm mode seems to be much better at handling this.
   + Do health check of the node via api call.  Kublets take more time to join the cluster than Swarm nodes.  Some type of sanity check / health check would be nice.

@YujiOshima

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,10 @@
 InfraKit
 ========
 
+[![CircleCI](https://circleci.com/gh/docker/infrakit/tree/master.svg?style=svg&circle-token=50d2063f283f98b7d94746416c979af3102275b5)](https://circleci.com/gh/docker/infrakit/tree/master)
+<!--
 [![Circle CI](https://circleci.com/gh/docker/infrakit.png?style=shield&circle-token=50d2063f283f98b7d94746416c979af3102275b5)](https://circleci.com/gh/docker/infrakit)
+-->
 [![Go Report Card](https://goreportcard.com/badge/github.com/docker/infrakit)](https://goreportcard.com/report/github.com/docker/infrakit)
 <!--
 [![codecov.io](https://codecov.io/github/docker/infrakit/coverage.svg?branch=master&token=z08ZKeIJfA)](https://codecov.io/github/docker/infrakit?branch=master)
--- a/README.md
+++ b/README.md
@@ -1,7 +1,8 @@
 InfraKit
 ========
-
+<!--
 [![CircleCI](https://circleci.com/gh/docker/infrakit/tree/master.svg?style=svg&circle-token=50d2063f283f98b7d94746416c979af3102275b5)](https://circleci.com/gh/docker/infrakit/tree/master)
+-->
 <!--
 [![Circle CI](https://circleci.com/gh/docker/infrakit.png?style=shield&circle-token=50d2063f283f98b7d94746416c979af3102275b5)](https://circleci.com/gh/docker/infrakit)
 -->
--- a/README.md
+++ b/README.md
@@ -1,8 +1,6 @@
 InfraKit
 ========
-<!--
 [![CircleCI](https://circleci.com/gh/docker/infrakit/tree/master.svg?style=svg&circle-token=50d2063f283f98b7d94746416c979af3102275b5)](https://circleci.com/gh/docker/infrakit/tree/master)
--->
 <!--
 [![Circle CI](https://circleci.com/gh/docker/infrakit.png?style=shield&circle-token=50d2063f283f98b7d94746416c979af3102275b5)](https://circleci.com/gh/docker/infrakit)
 -->
--- a/cmd/infrakit/util/init/init.go
+++ b/cmd/infrakit/util/init/init.go
@@ -96,9 +96,12 @@ func Command(plugins func() discovery.Plugins) *cobra.Command {
 
 	cmd.Flags().AddFlagSet(services.ProcessTemplateFlags)
 	groupID := cmd.Flags().String("group-id", "", "Group ID")
+	sequence := cmd.Flags().Uint("sequence", 0, "Sequence in the group")
 
 	configURL := cmd.Flags().String("config-url", "", "URL for the startup configs")
 	starts := cmd.Flags().StringSlice("start", []string{}, "start spec for plugin just like infrakit plugin start")
+
+	debug := cmd.Flags().Bool("debug", false, "True to debug with lots of traces")
 
 	cmd.RunE = func(c *cobra.Command, args []string) error {
 
@@ -107,12 +110,14 @@ func Command(plugins func() discovery.Plugins) *cobra.Command {
 			os.Exit(1)
 		}
 
-		logutil.Configure(&logutil.Options{
-			Level:    3,
-			Stdout:   false,
-			Format:   "term",
-			CallFunc: true,
-		})
+		if !*debug {
+			logutil.Configure(&logutil.Options{
+				Level:    3,
+				Stdout:   false,
+				Format:   "term",
+				CallFunc: true,
+			})
+		}
 
 		pluginManager, err := startPlugins(plugins, services, *configURL, *starts)
 		if err != nil {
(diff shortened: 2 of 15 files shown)


YujiOshima:
Great ! I will take a look!

README.md
```diff
@@ -1,7 +1,11 @@
 InfraKit
 ========
-
+<!--
```
AkihiroSuda: ?

chungers: The CircleCI badge all of a sudden stops working and I can't figure out why it's stuck at "failing" even though master builds successfully.  Any ideas?  I am not thrilled with CircleCI anymore.  Time to look at other options.

chungers:
Example here: https://github.com/chungers/examples/tree/master/latest/kube

YujiOshima:
I still can't have try `infrakit util init`, I would be grateful if you add more detailed readme for the example..

One of main point of this PR, the control plane node looks to make Infrakit HA and not related to k8s plugin, right?
I'm little bit confused what components are running where.
My understanding is below.
when I want 3 k8s cluster.
At the bootstrapping node, Infrakit in a docker container will be run first.

bootstrapping node : `Infrakit in docker`, `docker swarm manager`
nodeA : `docker swarm manager`, `k8s master`
nodeB : `docker swarm manager`, `k8s worker`
nodeC : `k8s worker`

Right?

chungers:
@YujiOshima 

In your example.... you want a 3 node cluster...  you will have exactly 3 nodes.  No extra node to run Docker swarm + Infrakit (where Infrakit uses Swarm mode to achieve HA).  You will have a group of managers with 3 logical IDs: 10.20.100.101, 10.20.100.102, 10.20.100.103.  In the Kube flavor plugin there's a `ControlPlane` field and we are going to set that to `[ "10.20.100.101" ]` to indicate that of these three logical IDs, only `10.20.100.101` is to be a member of the control plane for Kubernetes.  This is the single master case (the only one we support for now).  So the nodes:

  1. 10.20.100.101 - is the 'bootstrap node' which will have `infrakit util init` in its cloud-init script.  This corresponds to the [example here](https://github.com/chungers/examples/blob/master/latest/kube/aws/vpc.cfn#L373) (there's only 1 node in this AWS CloudFormation script).  Running on this node:
    + Docker engine (can be the latest from get.docker.com)
    + Running as leader (swarm init)
    + Has Infrakit running as container.  This is the leader node in infrakit HA setup
    + This also becomes the kube manager via the `ControlPlane` field in the config JSON for the Kube flavor plugin.  The `ControlPlane` field tells the flavor which nodes it should generate the manager init.  Previously the flavor will error out if more than one node is to be the master.  I am removing this for now because the plan is to support a multi-master HA setup where each kube master is also running on infrakit + docker manager.
  2. 10.20.100.102 - is the second infrakit manager node in this HA setup:
     + It's the second of the 3 swarm mode managers.
     + It's running infrakit in a container, but this infrakit container is not active because this node is not a leader of the swarm managers.
     + Because the `ControlPlane` field has only one member in the slice, this node, even though it's a infrakit / swarm mode manager node, it is considered by the Kube flavor plugin to be a 'data plane' node; therefore, it will join the kube cluster as a worker.
  3. 10.20.100.103 - same logic:
     + It's the 3rd node of the swarm / infrakit managers
     + It's not an active infrakit node because this swarm node is not a leader.
     + Because this node is not specified in the `ControlPlane` field, it's a data plane node for kube.  It's therefore a worker in the kube cluster.

If you have a group of workers, then they are all swarm nodes/ workers and kube workers because any additional nodes which are not specified in the `ControlPlane` slice will be considered a worker node.

This set up allows us to build a kube cluster on top of a swarm simply by a "mix-in" of configuration via a combo flavor and the kube flavor.  Ideally every infrakit manager node is also a kube master, but for now we don't support multi-master kube setup (I am not sure how kubeadm handles this), so `ControlPlane` is used to indicate which manager node should be the kube master, while any nodes that are not in the list will be a regular data plane worker node.  In this set up, there isn't a N+1 node for management as the previous kube flavor plugin assumes.  We can build a self-starting / bootstrapping kube cluster this way without waste of an extra node.

I hope this makes sense.  Let me know if I am not clear.

As for usage of `infrakit util init` -- think of it as a command line that does the following:
   + accepts input of a groups.json
   + given the flag `--group-id`, look for the group spec in the groups.json.
   + if `--sequence` is provided, use the sequence as in index in the `LogicalIDs` of the group spec (if LogicalIDs are specified).
   + Generate a request to call the `Prepare` method of the flavor plugin for this group spec (in this example, `combo`).  Essentially this is what the group controller would do when provisioning the first node.
   + Take the response from the flavor plugin, extract the `Init` field because this has all the complex template/ scripting for how to init the node.  Apply any template variable interpolation as necessary (hence the `--var` flags in the example to set params like cluster name).
   + Dump the final Init script out to `os.Stdout`.

All this could have been done if I had infrakit running as server and just invoke the flavor plugin's `prepare` verb from the CLI, but this command just simplifies the start up of the plugins and parsing of groups.json.

Anyway, so the Init script of the first node is not printed to stdout.  We simply pipe it to `sh` in that first node's cloud-init script.  So the first node comes up using the same set of groups.json, templates, etc. as the 2nd and 3rd nodes in the group, etc.   

Makes sense?

YujiOshima:
@chungers Thank you very much for your kind instruction.
I understood the system constitution and what you want to do in this PR.
At DockerConEU, I want to discuss about next step of kube support as you listed above!

PR #723, opened 2017-10-12
Handle panics in tf provider apply goroutine

kaufers:
When the terraform files are on shared NFS we are seeing random panics in the `handleFiles` function; this function is executing in a goroutine that should always be running.

Until we can get to the bottom of what is ultimately causing these, this commit will recover from the `panic` and keep the goroutine alive.

### DIFF
--- a/pkg/provider/terraform/instance/apply.go
+++ b/pkg/provider/terraform/instance/apply.go
@@ -2,10 +2,12 @@ package instance
 
 import (
 	"bufio"
+	"fmt"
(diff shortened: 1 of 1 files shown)


kaufers:
Turns out that the `panic` was due to a `nil` reference to the `os.FileInfo` interface; when this is `nil` there is also an `error` passed in. We should be able to ignore the file entry when `err != nil`.

I will close this PR and push up another to address this.

PR #731, opened 2017-10-26
[WIP] HPE OneView Native InfraKit plugin

thebsdbox:
This is the re-written HPE OneView plugin for InfraKit, can be merged.

Done:
- [x] Needs moving `inproc` 
- [x] Needs moving to the correct logging pkg
- [x] Vendoring needs updating.

To do at a later Date:
- [ ] Some discussion around best practice for the group plugin timeout when `provision` method is called.
- [ ]  Move from an internal state manager, to the InfraKit FSM pkg.


pkg/provider/oneview/main.go
```diff
@@ -0,0 +1,53 @@
+package main
+
+import (
+	"os"
+
+	log "github.com/Sirupsen/logrus"
```
chungers: We're moving away from logrus.  The new logging system does call stacks, line numbers, filter by matches, and levels.  

Use 
```
import (
  logutil "github.com/docker/infrakit/pkg/log"
)

var log = logutil.New("module", "instance/oneview")
var debugV = logutil.V(200)  // 100-500 are for typical debug levels, > 500 for highly repetitive logs (e.g. from polling)
```
Then use `log.Info`, `log.Debug` as needed.  You can add a verbose level to the debug log like so:
```
log.Debug("this is log header/message", "field1", "value1", "field2", "value2", "V", debugV)
```
Then in the CLI:
```
infrakit .... --log 5 --log-debug-V 500 --log-debug-match module=instance/oneview
```
will log only DEBUG statements from your package (module=instance/oneview) and any thing with V < 500.

thebsdbox: Cool, i'll move over to that logging package and fix the vending.

Issue #733, opened 2017-10-27
Command-line parsing is non-deterministic


talex5:
I started the manager plugin with:

```infrakit plugin start manager```

Then doing `infrakit group ls -h` gives different output at different times, e.g.

```
$ infrakit group ls -h


List groups

Usage:
  infrakit group ls [flags]

Flags:
  -q, --quiet   Print rows without column headers

Global Flags:
      --alsologtostderr                  log to standard error as well as files
      --httptest.serve string            if non-empty, httptest.NewServer serves on this address and blocks
      --log int                          log level (default 4)
      --log-caller                       include caller function (default true)
      --log-debug-V int                  log debug verbosity level. 0=logs all
      --log-debug-match stringSlice      debug mode only -- select records with any of the k=v pairs
      --log-debug-match-exclude          True to exclude; otherwise only include matches
      --log-format string                log format: logfmt|term|json (default "term")
      --log-stack                        include caller stack
      --log-stdout                       log to stdout
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory
      --logtostderr                      log to standard error instead of files
      --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
  -v, --v Level                          log level for V logs
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
```
vs
```
$ infrakit group ls -h


List metadata

Usage:
  infrakit group ls [flags]

Flags:
  -a, --all     Find all under the paths given
  -l, --long    Print full path
  -q, --quick   True to turn off headers, etc.

Global Flags:
      --alsologtostderr                  log to standard error as well as files
      --httptest.serve string            if non-empty, httptest.NewServer serves on this address and blocks
      --log int                          log level (default 4)
      --log-caller                       include caller function (default true)
      --log-debug-V int                  log debug verbosity level. 0=logs all
      --log-debug-match stringSlice      debug mode only -- select records with any of the k=v pairs
      --log-debug-match-exclude          True to exclude; otherwise only include matches
      --log-format string                log format: logfmt|term|json (default "term")
      --log-stack                        include caller stack
      --log-stdout                       log to stdout
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory
      --logtostderr                      log to standard error instead of files
[...]

PR #734, opened 2017-10-27
Update to the ls vars to prevent command clash

FrenchBen:
Fixes https://github.com/docker/infrakit/issues/733 by preventing var name collision


chungers:
I think the fix is changing the "use" of cobra command, not renaming a local variable inside functions.  In any case, #736 should fix the problem by changing `ls` to `keys`.

PR #745, opened 2017-11-03
IBM Cloud loadbalancer implementation

ericvn:
Signed-off-by: Eric Van Norman <ericvn@us.ibm.com>

### DIFF
--- a/pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
+++ b/pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
@@ -1,0 +1,466 @@
+package ibmcloud
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	logutil "github.com/docker/infrakit/pkg/log"
+	"github.com/docker/infrakit/pkg/spi/instance"
+	"github.com/docker/infrakit/pkg/spi/loadbalancer"
+	"github.com/softlayer/softlayer-go/datatypes"
+	"github.com/softlayer/softlayer-go/services"
+	"github.com/softlayer/softlayer-go/session"
+)
+
+// For logging
+var lbLogger = logutil.New("module", "ibmcloud/loadbalancer")
+
+const (
+	debugV           = logutil.V(100)
+	updateRetryCount = 30
+	updateRetryTime  = 100 * time.Millisecond
+)
+
+// Result string
+type lbResult string
+
+// LBOptions are the configuration parameters for the LB provisioner.
+type LBOptions struct {
+	uuid string
+}
+
+// ibmcloudlb captures the options for starting up the plugin.
+type ibmcloudlb struct {
+	// name
+	name string
+
+	// UUID of the IBM Cloud load balancer backing this
+	uuid string
+
+	// mutex to prevent overlaps of calls. May not be needed.
+	lock sync.Mutex
+
+	// Softlayer user name
+	softlayerUsername string
+
+	// Softlayer API Key
+	softlayerAPIKey string
+
+	// IAM endpoint
+	iamEndpoint string
+
+	// Softlayer end point URL
+	softLayerEndpointURL string
+
+	// SoftlayerXMLRPCEndpoint endpoint
+	softlayerXMLRPCEndpoint string
+
+	// retry count for API calls
+	// Unexposed in the schema at this point as they are used only during session creation for a few calls
+	// When sdk implements it we an expose them for expected behaviour
+	// https://github.com/softlayer/softlayer-go/issues/41
+	retryCount int
+
+	// Constant retry delay for API calls
+	retryDelay time.Duration
+
+	// Session
+	session *session.Session
+
+	// Loadbalancer service
+	lbService services.Network_LBaaS_LoadBalancer
+
+	// Loadbalancer Listener service
+	lbListenerService services.Network_LBaaS_Listener
+
+	// Loadbalancer Member service
+	lbMemberService services.Network_LBaaS_Member
+
+	// Loadbalancer health Monitor  service
+	lbHealthMonitorService services.Network_LBaaS_HealthMonitor
+
+	// SSL Certificate service
+	certService services.Security_Certificate
+}
+
+// NewIBMCloudLBPlugin returns a L4 loadbalancer backed by the IBM cloud
+func NewIBMCloudLBPlugin(username, apikey, lbName, lbUUID string) (loadbalancer.L4, error) {
+	lbLogger.Debug("NewIBMCloudLBPlugin", "V", debugV, "Name", lbName)
+	lb := &ibmcloudlb{
+		softlayerUsername: username,
+		softlayerAPIKey:   apikey,
+		name:              lbName,
+		uuid:              lbUUID,
+	}
+
+	// Hardcoded IBM Cloud endpoints
+	lb.iamEndpoint = "https://iam.ng.bluemix.net"
+	lb.softlayerXMLRPCEndpoint = "https://api.softlayer.com/xmlrpc/v3"
+	lb.softLayerEndpointURL = "https://api.softlayer.com/rest/v3.1"
+
+	// Hardcoded retry logic for the IBM Cloud calls
+	lb.retryCount = 10
+	lb.retryDelay = 20 * time.Millisecond
+
+	// Create the session object and get the services object once
+	lb.session = session.New(lb.softlayerUsername, lb.softlayerAPIKey, lb.softLayerEndpointURL)
+	lb.lbService = services.GetNetworkLBaaSLoadBalancerService(lb.session)
+	lb.certService = services.GetSecurityCertificateService(lb.session)
+	lb.lbListenerService = services.GetNetworkLBaaSListenerService(lb.session)
+	lb.lbMemberService = services.GetNetworkLBaaSMemberService(lb.session)
+	lb.lbHealthMonitorService = services.GetNetworkLBaaSHealthMonitorService(lb.session)
+
+	return lb, nil
+}
+
+func (r lbResult) String() string {
+	return string(r)
+}
+
+// stringPtr returns a pointer to the string value passed in.
+func stringPtr(v string) *string {
+	return &v
+}
+
+// intPtr returns a pointer to the int value passed in.
+func intPtr(v int) *int {
+	return &v
+}
+
+// max returns the max int of two passed in ints.
+func max(x, y int) int {
+	if x > y {
+		return x
+	}
+	return y
+}
+
+// Name is the name of the load balancer
+func (l *ibmcloudlb) Name() string {
+	lbLogger.Debug("Name", "V", debugV)
+	return l.name
+}
+
+// Routes lists all known routes.
+func (l *ibmcloudlb) Routes() ([]loadbalancer.Route, error) {
+	lbLogger.Debug("Routes", "V", debugV)
+	l.lock.Lock()
+	defer l.lock.Unlock()
+
+	// Output array
+	out := []loadbalancer.Route{}
+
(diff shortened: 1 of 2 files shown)


docs/controller/ingress/example.yml
```diff
@@ -16,9 +16,9 @@
 # Verify that the route has been added in the simulator/lb1:
 # infrakit simulator/lb1 routes ls
 #
-# FRONTEND PORT    PROTOCOL   BACKEND PORT  CERT
-# 7777             TCP        7777  <-- from the swarm service
-# 80               http       8080  <-- from the static route in the config
+# FRONTEND PORT    FRONTEND PROTOCOL      BACKEND PORT      BACKEND PROTOCOL        CERT
```
kaufers: Why are these changes here? Is this file out of date? Can you put something in the commit message explaining why this is updated? It seems like that the `Protocol` case is the main delta here (`http` -> `HTTP`).

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
```diff
@@ -0,0 +1,466 @@
...
+
+	// Iterate over the list of members to add checking if they already exist in the loadbalancer.
+	// If not, add member.
+	for _, id := range ids {
+		ip := string(id)
+		ipFound := false
+		for _, member := range lb.Members {
+			if *member.Address == ip {
+				ipFound = true
+				lbLogger.Error("RegisterBackends", "name", l.name, "V", debugV, "backend already exists", ip)
```
kaufers: Why is this an error? Won't this be hit if there is a rolling update and now we have a new member?

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
```diff
@@ -0,0 +1,466 @@
...
+	// Get current list of backend members
+	lb, err := l.lbService.Mask("members").GetLoadBalancer(&l.uuid)
+	if err != nil {
+		return nil, err
+	}
+
+	// Iterate over the list of members to add checking if they already exist in the loadbalancer.
+	// If not, add member.
+	for _, id := range ids {
+		ip := string(id)
```
kaufers: So the `ids` that are passed in are actually IPv4 addresses? If so, can you add a comment that explains this?

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
```diff
@@ -0,0 +1,466 @@
...
+		attempt := 0
+		for {
+			attempt++
+			_, err = l.lbMemberService.AddLoadBalancerMembers(&l.uuid, membersToAdd)
+			if err == nil {
+				break
+			}
+			// Check for the lb is in state UPDATE_PENDING. (HTTP 500)
+			if strings.Contains(err.Error(), "UPDATE_PENDING") && attempt <= updateRetryCount {
+				time.Sleep(updateRetryTime)
```
kaufers: Can we add in some logging here that exposes this error case? I like that we are handling it but it would be nice to get visibility into how often this happens.

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
```diff
@@ -0,0 +1,466 @@
...
+		ip := string(id)
+		for _, member := range lb.Members {
+			if *member.Address == ip {
+				membersToRemove = append(membersToRemove, *member.Uuid)
+				ipFound = true
+				break
+			}
+		}
+		if ipFound == false {
+			lbLogger.Error("DeregisterBackends", "name", l.name, "V", debugV, "backend not found", ip)
```
kaufers: Again why is this an error? Shouldn't be the normal case when the group member changes (like on a rolling update)?

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
```diff
@@ -0,0 +1,466 @@
...
+		attempt := 0
+		for {
+			attempt++
+			_, err = l.lbMemberService.DeleteLoadBalancerMembers(&l.uuid, membersToRemove)
+			if err == nil {
+				break
+			}
+			// Check for the lb is in state UPDATE_PENDING. (HTTP 500)
+			if strings.Contains(err.Error(), "UPDATE_PENDING") && attempt <= updateRetryCount {
+				time.Sleep(updateRetryTime)
```
kaufers: Same comment on the logging.

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
```diff
@@ -0,0 +1,466 @@
...
+				return nil, err
+			}
+		}
+	}
+
+	lbLogger.Debug("DeregisterBackends", "name", l.name, "V", debugV, "backends", membersToRemove)
+	return lbResult("deregister"), nil
+}
+
+// Backends returns a list of backends
```
kaufers: Can we explain that we are using the IP address as a common correlation key? Even though this func is returning an `[]instance.ID`, the value is simply the IP address of each member, right?

pkg/run/v0/ibmcloud/ibmcloud.go
```diff
@@ -39,11 +49,38 @@ type VolumeAuth struct {
 	VolumeID int
 }
 
+// LB is custom type for IBM Cloud loadbalancer definition
+type LB struct {
+	// LBName is the name of the IBM Cloud loadbalancer
+	LBName string
+
+	// LBUUID is the UUID of the IBM cloud loadbalancer
```
kaufers: s/IBM cloud/IBM Cloud

pkg/run/v0/ibmcloud/ibmcloud.go
```diff
@@ -76,11 +114,29 @@ func Run(scope scope.Scope, name plugin.Name,
 		authInstPlugin = ibmcloud_auth_inst.NewVolumeAuthPlugin(options.Username, options.APIKey, options.VolumeAuth.VolumeID)
 	}
 
+	l4Map := map[string]loadbalancer.L4{}
+	for _, lb := range options.LBs {
+		var lbPlugin loadbalancer.L4
+		lbPlugin, err = ibmcloud_loadbalancer.NewIBMCloudLBPlugin(options.Username, options.APIKey, lb.LBName, lb.LBUUID)
+		if err != nil {
+			return
```
kaufers: Should we log an error here?

docs/controller/ingress/example.yml
ericvn: They've simply been sitting in my branch waiting to be put in as David said to put them in with my next PR. I could certainly push them in another PR if that is warranted.

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
ericvn: I though I put this error message into the code pulling from either the simulator or AWS load balancer code, but can't find that now. This code logs an error message that the backend was already registered, but doesn't change any behavior. I'm fine with removing this check entirely, or downgrading the message to a warning/info/debug.

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
ericvn: Added into the function remarks.

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
ericvn: Done in the several places we have this code.

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
ericvn: Similar to the above comment.

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
ericvn: Done.

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
ericvn: Correct.

ericvn: I'll add a bigger remark that we are using the private IP address of the worker instead of the worker name. This was because there wasn't a field to map worker name to ip within the LBaaS APIs.

pkg/run/v0/ibmcloud/ibmcloud.go
ericvn: I find the logging inconsistent across plugins, so adding it here.

pkg/run/v0/ibmcloud/ibmcloud.go
ericvn: Done

docs/controller/ingress/example.yml
```diff
@@ -58,9 +58,9 @@ properties:
     # Here we have a static route that is always present.
     Routes:
       - LoadBalancerPort: 80
-        LoadBalancerProtocol: https
+        LoadBalancerProtocol: HTTP
```
chungers: Did this become case sensitive at some point?   Curious why you'd change it to all caps.

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
```diff
@@ -0,0 +1,488 @@
...
+		}
+		if len(cert) == 1 {
+			certID = *cert[0].Id
+		} else {
+			return nil, fmt.Errorf("Cannot identify an unique certificate with common name '%s'", *route.Certificate)
+		}
+	}
+
+	// InfraKit doesn't define a method, so default to round robin
+	lbMethod := "ROUNDROBIN"
```
chungers: Is this something that we can add to the `Route` struct?  Different L4 implementations will have different strings but if this is just some opaque value that's passed on to the L4 then we can avoid this hardcoding here.

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
```diff
@@ -0,0 +1,488 @@
...
+	lbLogger.Debug("NewIBMCloudLBPlugin", "V", debugV, "Name", lbName)
+	lb := &ibmcloudlb{
+		softlayerUsername: username,
+		softlayerAPIKey:   apikey,
+		name:              lbName,
+		uuid:              lbUUID,
+	}
+
+	// Hardcoded IBM Cloud endpoints
+	lb.iamEndpoint = "https://iam.ng.bluemix.net"
```
chungers: It might be useful to define a few exported / public constants and not have these values buried in here.

docs/controller/ingress/example.yml
ericvn: Let me go back and look at this one as I made it a while ago. I recall at the time that the check between what is in the yml and what the Routes API returned for protocol was case sensitive, and this might have been fixed. I moved to the Protocol case in protocol.go.

docs/controller/ingress/example.yml
chungers: Let's remove changes to this file in this PR...

ericvn: Removed.

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
ericvn: The specification of the method could be added. For information, the IBM LBaaS LB supports 3 methods: Round Robin, Weighted Round Robin (would also require the addition of a weight on each backend), and Least Connection.

pkg/provider/ibmcloud/plugin/loadbalancer/lb.go
ericvn: Done.

PR #759, opened 2017-11-09
VMware vSphere logging and stability fixes.

thebsdbox:
Removed the old style of logging, also updated comments and implemented
a State machine for systems with slow storage. This should allow
reliable provisioning regardless of the usage of VAAI/VAPI, closing one
of the issues and bringing the plugin inline. (FSM will need changing
later on)

### DIFF
--- a/pkg/provider/vsphere/plugin.go
+++ b/pkg/provider/vsphere/plugin.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
-	"os"
 	"time"
 
 	"github.com/docker/infrakit/pkg/spi"
@@ -35,9 +34,9 @@ func init() {
 
 //miniFSM for managing the provisioning -> provisioned state
 type provisioningFSM struct {
-	countdown    int64             // ideally will be a counter of minutes / seconds
+	timer        time.Time         // ideally will be a counter of minutes / seconds
 	tags         map[string]string // tags that will be passed back per a describe function
-	instanceName string            // name that we will use as a lookup to the actual backend that is privisioning
+	instanceName string            // name that we will use as a lookup to the actual backend that is provisioning
 }
 
 // NewInstancePlugin will take the cmdline/env configuration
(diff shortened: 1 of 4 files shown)


pkg/provider/vsphere/plugin.go
```diff
@@ -48,7 +60,8 @@ func NewVSphereInstancePlugin(vc *vCenter, ignoreOnDestroy bool) instance.Plugin
 	internals, err := vCenterConnect(vc)
 	if err != nil {
 		// Exit with an error if we can't connect to vCenter (no point continuing)
-		log.Fatalf("%v\n", err)
+		log.Crit("Error connecting to vCenter", "response", err.Error())
+		os.Exit(-1)
```
chungers: I think it's better to let this error bubble up the stack, rather than just exiting here.  This will make it possible for the upper layer to do things in the future like retries without you having to do it yourself here.

pkg/provider/vsphere/plugin.go
```diff
@@ -104,15 +117,15 @@ func (p *plugin) Provision(spec instance.Spec) (*instance.ID, error) {
 		}
 	}
 
-	newInstance, err := parseParameters(properties, p)
+	newVM, err := parseParameters(properties, p)
 	if err != nil {
-		log.Errorf("Error: \n%v", err)
+		log.Error("Problems Whilst Parsting", "Error", err)
```
chungers: Sorry a nit pick here.  We haven't had a convention for logging, so maybe now is the time to start...  All the logging statements elsewhere use `"err"` as a key for error.  Would you mind changing to that?

pkg/provider/vsphere/vsphere.go
```diff
@@ -151,15 +149,15 @@ func parseParameters(properties map[string]interface{}, p *plugin) (vmInstance,
 		return newInstance, errors.New("Property 'Datastore' must be set")
 	}
 	*p.vC.dsName = properties["Datastore"].(string)
-	log.Debugf("Datastore set to %s", *p.vC.dsName)
+	log.Debug(fmt.Sprintf("Datastore set to %s", *p.vC.dsName))
```
chungers: The convention for log15 (the library we use) is 
```
log.Debug("a general message", "key1", value1, "key2", value2)
```
So this log statement should be rewritten to
```
log.Debug("Setting datastore", "dataStore", *p.vC.dsName)
```

pkg/provider/vsphere/plugin.go
thebsdbox: Wasn’t sure of the code path.  Didn’t seem to make sense replying with a plugin struct when the config was incorrect. I was also wondering why there wasn’t a fatal option as part of the logging. I need to re-consider the auth strategy for this part of the plugin moving forward.

pkg/provider/vsphere/plugin.go
thebsdbox: Sure, I did find the logging a little confusing (lack of fatal etc.) a convention would help
 :-)

PR #760, opened 2017-11-10
Check for recent file deltas before running tf apply

kaufers:
When multiple `Provision` requests come in we really want to process them all at once. This commit adds support to look at the file timestamps and only run `terraform apply` if there have been no
updates (ie, no `Provision` requests) in the last 3 seconds.

### DIFF
--- a/pkg/provider/terraform/instance/apply.go
+++ b/pkg/provider/terraform/instance/apply.go
@@ -2,6 +2,7 @@ package instance
 
 import (
 	"bufio"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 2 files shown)


pkg/provider/terraform/instance/apply.go
```diff
@@ -136,6 +149,53 @@ type tfFuncs struct {
...
+				}
+			}
+			return nil
+		},
+	)
+	if err != nil {
+		return false, err
+	}
+	if !modTime.IsZero() {
+		now := time.Now()
```
chungers: This is really a clever idea.  Essentially we are introducing the idea of check points into the processing of tf config files.

I am wondering here with this, will we be missing updates when the plugin shuts down and comes back up?  What if we write the timestamp of last apply and use that as a check for all the modifications since that checkpoint?

PR #761, opened 2017-11-10
Update terraform file system lock to use a mutex

kaufers:
The current file system lock was intended to prevent multiple instances of the instance plugin from reading/writing the file system; **it does not serialize operations for the current instance**. The file lock is not necessary since the group controller only runs on the leader and the tf `apply` goroutine only runs `terraform ...` commands on the manager. Moreover, if the leader dies (or looses network connectivity) while it has the process lock then we are deadlocked.

This causes issues if multiple `Provision` requests come in (and quorum groups do not serialize `Provision` requests) because the same `instance-<timestamp>` files may be used (then then the last request overrides the file).

The fix is to use a mutex instead of a file lock.

### DIFF
--- a/pkg/provider/terraform/instance/apply.go
+++ b/pkg/provider/terraform/instance/apply.go
@@ -144,10 +144,7 @@ type tfFuncs struct {
 // Once these steps are done then "terraform apply" can execute without the
 // file system lock.
 func (p *plugin) handleFiles(fns tfFuncs) error {
-	if err := p.fsLock.TryLock(); err != nil {
-		log.Infof("In handleFiles, cannot acquire file lock")
-		return err
-	}
+	p.fsLock.Lock()
 	defer p.fsLock.Unlock()
 
 	// Refresh resources and get updated resources names
--- a/pkg/provider/terraform/instance/plugin.go
+++ b/pkg/provider/terraform/instance/plugin.go
@@ -23,7 +23,6 @@ import (
 	"github.com/docker/infrakit/pkg/template"
 	"github.com/docker/infrakit/pkg/types"
 	"github.com/docker/infrakit/pkg/util/exec"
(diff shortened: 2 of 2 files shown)


kaufers:
@chungers 

> the filesystem lock could just be replaced by a mutex as long as each terraform plugin has its own dedicated folder

What do you mean by this? Isn't the group controller going to ensure that the instance APIs are only executed on the leader? I thought that that file system lock was created in order to prevent multiple plugins from colliding -- but that shouldn't never happen since they only run on the leader.

> As long as we are removing the filesystem lock, would you mind update the vendor.conf?

Sure

chungers:
>> the filesystem lock could just be replaced by a mutex as long as each terraform plugin has its own dedicated folder

> What do you mean by this? Isn't the group controller going to ensure that the instance APIs are only executed on the leader? I thought that that file system lock was created in order to prevent multiple plugins from colliding -- but that shouldn't never happen since they only run on the leader.

This is because it's actually possible to do this:

```
infrakit plugin start terraform:nodes terraform:networks terraform:resources
```
This will actually start up 3 terraform plugins with different names and the assumptions here is that the user can have each terraform plugin instance manage different types of resources -- maybe the `nodes` one is for group instances and `resources` just create some resources via enrollment controller, etc.

I don't think it's unreasonable to assume that these three terraform plugin instances are unrelated and are therefore required to store files in separate directories.  The lock file was originally trying to prevent this, but lockfile is such a weak contract to begin with that I think the problem is better solved by reasonable deployment assumptions and simpler implementation like mutexs and making sure controllers serialize operations where necessary.

kaufers:
@chungers Yep, in that case, the user could use the `Dir` `Option` to configure each instance with a unique directory for it's terraform files.

https://github.com/docker/infrakit/tree/master/pkg/provider/terraform/instance#runtime-options

PR #768, opened 2017-11-16
Number of additional features

thebsdbox:
- Logging fixes
- Additional Networking scheme, allowing networking configuration to be applied per build
- Fix to handle older RHEL/CentOS sudo issues.
- README update for updated environment variables.

### DIFF
--- a/pkg/x/vmwscript/vm_provision.go
+++ b/pkg/x/vmwscript/vm_provision.go
@@ -185,7 +185,8 @@ func RunTasks(ctx context.Context, client *govmomi.Client) {
 			// Check if a networking configuration exists
 			if task.Task.Network != nil {
 				// Determine the distribution to configure the networking
-				if task.Task.Network.Distro == "centos" {
+				switch task.Task.Network.Distro {
+				case "centos": // Set up networking for the CentOS distribution
 					setCentosNetwork(ctx, client, newVM, auth, task.Task.Network)
 					log.Info("Restarting Virtual Machine, and waiting for tools")
 					newVM.RebootGuest(ctx)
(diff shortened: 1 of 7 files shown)


pkg/x/vmwscript/vm_provision.go
```diff
@@ -185,13 +185,43 @@ func RunTasks(ctx context.Context, client *govmomi.Client) {
 				Password: *vm.VMTemplateAuth.Password,
 			}
 
+			// Check if a networking configuration exists
+			if task.Task.Network != nil {
+				// Determine the distribution to configure the networking
+				if task.Task.Network.Distro == "centos" {
```
chungers: How about a switch here:

```
switch task.Task.Network.Distro {
   case "centos" :  // call a function here.
   default :  // return some 'unsupported' error.
}
```

This makes it clear that the Distro field will be something important to add more cases in the future...

pkg/x/vmwscript/vm_provision.go
```diff
@@ -392,10 +425,90 @@ func watchPid(ctx context.Context, client *govmomi.Client, vm *object.VirtualMac
...
+				log.Warn("Process no longer watched, VMware Tools may have been restarted")
 				break
 			}
 		}
 	}
 	return nil
 }
+
+// setCentosNetwork - CentOS only networking
+func setCentosNetwork(ctx context.Context, client *govmomi.Client, vm *object.VirtualMachine, auth *types.NamePasswordAuthentication, netConfig *NetworkConfig) error {
```
chungers: Just a comment... nothing for you to do at the moment...

We should look at networking setups for a couple of other platforms and come up with a common interface for this.

thebsdbox: Yeah that's the plan. I just need to install some more distros and then I can come up with some sane way of presenting a network configuration back to the end user regardless of underlying distro. Not sure if it would be worth doing the same for things like firewall rules as well, but then it depends how far down the hole we want to go. The main reason for this is just to bootstrap a single instance that InfraKit can then make use of to deploy and scale etc.

PR #773, opened 2017-11-21
VMWScript engine code updates

chungers:
This PR suggests a number of changes to the VMWScript engine in preparation of deeper integration with infrakit as a playbook backend:

  + Avoid the use of package-scoped, static variables (`plan`, `commandCounter`, etc.) which will not be thread-safe if multiple VMWScript engines are running in the same process.
  + Moved these states into the `DeploymentPlan` struct so that these are instance variables which track the state of the running engine instance.

Next step after this is to embed this engine as a Playbook backend (see pkg/cli/backend) so that users can create custom VMWScripts with added functionality of  templating and including other scripts and command line flags integration with the CLI.

### DIFF
--- a/cmd/infrakit/x/vmwscript.go
+++ b/cmd/infrakit/x/vmwscript.go
@@ -16,8 +16,6 @@ var cmdResults = map[string]string{}
 var debugV = logutil.V(200) // 100-500 are for typical debug levels, > 500 for highly repetitive logs (e.g. from polling)
 
 func vmwscriptCommand() *cobra.Command {
-	vmwscript.InitDeployment()
-	vm := vmwscript.VMwareConfig() //Pull VMware configuration from JSON
 
 	var vc, dc, ds, nn, vh, gu, gp *string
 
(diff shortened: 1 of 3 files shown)


cmd/infrakit/x/vmwscript.go
```diff
@@ -33,76 +31,38 @@ func vmwscriptCommand() *cobra.Command {
...
-			}
-		}
+		// set the values from flags
+		plan.VMWConfig.VCenterURL = vc
+		plan.VMWConfig.DCName = dc
+		plan.VMWConfig.DSName = ds
+		plan.VMWConfig.NetworkName = nn
+		plan.VMWConfig.VSphereHost = vh
+		plan.VMWConfig.VMTemplateAuth.Username = gu
+		plan.VMWConfig.VMTemplateAuth.Password = gp
```
thebsdbox: Originally we checked the `VMWConfig.{X}` to see if it had been set from the unmarshalling of the JSON, if not then we set it to what was either a flag or env variable. This change results in the wipe of any configuration set from the JSON to what is either a flag or environment variable.

thebsdbox:
Also, just successfully tested `pkg/x/vmwscript/examples/swarm.json` with these changes.

cmd/infrakit/x/vmwscript.go
chungers: Ah sorry I missed that.  This should be easy to fix: we instantiate an empty Plan and the decode with the bytes from a file then call a Validate.  Will work on that right away.

chungers: I meant after instantiating the Plan we set its fields with flags values and then Validate.

PR #774, opened 2017-11-21
fix nil pointer in vmwscript cli

chungers:


thebsdbox:
Tested fine... Previously we would compare if alternative config both existed or was just empty i.e. (`""` or `nil`), this change means that the examples will wipe out flags or environment variables as they have `JSON` config that sets config to `""`. Ideally we need to update the examples at somepoint.

PR #780, opened 2017-11-28
Added Capability to run standalone commands against existing VM

thebsdbox:
This adds the capability to run commands on an ad-hoc approach against an existing virtual machine with three new flags (no script needs passing)

      `--vmcommand string`      A command passed as a string to be executed on the virtual machine specified with [--vmname]
      `--vmname string`         The name of an existing virtual machine to run a command against
      `--vmsudouser string`     A sudo user that the command will be executed
  

thebsdbox:
Added in some documentation. The long term goal of this is being able to pull things from a running virtual machine (like swarm keys from a UCP) and then use them with the vSphere plugin as it will spin out some worker instances.

PR #790, opened 2017-12-01
Map TCP to HTTPS if cert specified.

craigyam:
Switch from mapping TCP/TCP frontend/backend protocols to SSL/TCP when
specifying a cert instead to HTTPS/HTTP.

Signed-off-by: Craig Yamamoto <craigyam@us.ibm.com>


craigyam:
@chungers Fixed the tests.  Thanks!

Issue #783, opened 2017-11-29
Rolling updates of the manager nodes


kaufers:
We are running Infrakit on the swarm manager nodes and want to issue a rolling update of `managers` group.

**It appears that Infrakit is not waiting for each manager to complete before proceeding to the next node.**

Note that we are using block storage to backup `/var/lib/docker` and re-using the same IP address for each manager. This means that, during a rolling update of a manager node, the node is not removed from the swarm (so `docker node ls` still shows it).

A possible issue is that the `Destroy` is asynchronous and that the `DescribeInstances` nodes have no status attribute (like "provisioning" or "complete"). Here's what the flow looks like:

1. Group controller issues a `DescribeInstances` and gets instances with Logical ID values of `mgr1`, `mgr2`, `mgr3`
2. Group controller issues a `Destroy` on node with Logical ID of `mgr1` and then issues a `Provision` with a Logical ID of `mgr1`
3. The next `DescribeInstances` poll cycle from the group controller gets instances with Logical ID values of `mgr1`, `mgr2`, `mgr3`. In this case the "new" `mgr1` is not yet provisioned -- moreover, the "old" `mgr1` is probably still even online since the backend cloud VM destroy takes a minute or so to power down the VM (which means that the swarm node is still `ready` and `available`)
4. Group controller issues a `Destroy` on node with Logical ID of `mgr2` ... etc.

Do we need to return some "provisioning status" on the `DescribeInstances` so that the group controller can determine if a node still being provisioned or if it is complete?

chungers:
This is related to #782.  There are parts to this process for managers.   For managers, we assume parallelism of 1 which means one sequential thread of updates for this entire group.

  + Ordering of rolling update for managers
  + Ensuring proper replacement before continuing -- this includes health-check (e.g. new manager joins the swarm)
  + Transferring leadership *proactively* to a new updated manager node
  + Recording the operations so that when leadership is transferred, the new leader can continue this process correctly.

Some key issues:

### Order of Rolling Updates

We need to compute the correct ordering of the nodes before rolling update for certain types of groups...   for workers this is not essential but for managers this is critical in order to avoid scenario where the leader itself gets terminated.

[...]

PR #791, opened 2017-12-01
Handle manager failover during terraform provision

kaufers:
In the terraform instance plugin an orphaned resource `tf.json` file can exist for 2 different reasons:
1. Resource has been removed out-of-band from Infrakit, in this case the resource in the terraform state file will be removed during a `terraform refresh` operation.
2. Infrakit failed over to a different manager during a `terraform apply`, in this case the state file is **not** updated because the apply never finished (terraform only updates at the end of a provision -- nothing is updating _during_ a provision).

In order to differentiate between these 2 cases we need to get a list of existing resources **before** and **after** we execute `terraform refresh`.

When we detect a orphan, we can simply remove the file if the resource previously existed and is not removed (case 1 above). However, if the resource did not exist, then we need to:
1. Query the backend cloud for VM with the unique tags
2. If a VM is found then import it into terraform
3. If no VM is found then prune the `tf.json` file

This allows us to complete an in-progress provision when management fails over (or if the container is restarted for whatever reason during a provision).

The downside of this solution is that is requires platform-specific updates to perform the backend query; this commit contains the implementation for IBM Cloud only.

### DIFF
--- a/pkg/provider/terraform/instance/apply.go
+++ b/pkg/provider/terraform/instance/apply.go
@@ -7,15 +7,25 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/deckarep/golang-set"
 	manager_discovery "github.com/docker/infrakit/pkg/manager/discovery"
 	"github.com/docker/infrakit/pkg/types"
 	"github.com/docker/infrakit/pkg/util/exec"
 	"github.com/spf13/afero"
 )
+
+// TResourceFilenameProps contains the filename and the associated properties
+// for a specific resource
+type TResourceFilenameProps struct {
+	FileName  string
+	FileProps TResourceProperties
+}
 
 // terraformApply starts a goroutine that executes "terraform apply" at the
 // configured freqency; if the goroutine is already running then the sleeping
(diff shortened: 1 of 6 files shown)


pkg/provider/terraform/instance/apply.go
```diff
@@ -329,6 +353,146 @@ func (p *plugin) handleFiles(fns tfFuncs) error {
...
+		err := p.fs.Remove(path)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// getExistingResource queries the backend cloud to get the ID of the resource associated
+// with the given type, name, and properties
```
chungers: As discussed in #783 do you think storing the state of an update will obviate this - at least for the rolling update?   Presumably the manager will never fail over unexpectedly if we implemented safe ordering of managers to update and there will be state in raft that the new leader can read and resume.

chungers: Also if solutions for #783 is implemented then the case 2 as you describe would  never happen, right?  At least not intentionally...

kaufers: @chungers Yes, you are correct. If we failover to a new manager _before_ we start the rolling update then this should not be needed.

However, this PR fixes the generic case when the container dies/restarts _during_ a provision for any reason; for example, if Infrakit is deployed as a Docker service, then doing something like `docker service update --image <new-image>` would cause the associated containers to restart.

PR #817, opened 2017-12-26
Support for ssh

chungers:
This PR adds support for executing commands over ssh on remote hosts.

  + While we strongly prefer immutable infrastructure, ssh is sometimes required to manage the configurations of provisioned nodes (ansible style).
  + Support for basic password / private key and ssh agent auths
  + Integration with playbook as a backend choice
  + Backends can add additional flags that are applied for all playbooks using that backend (e.g. ssh remote host, user, password, etc.)
  + Support for multiple SSH remotes and execution of scripts in parallel.

Documentation:  See `pkg/cli/backend/ssh/README.md` on how to create playbooks using SSH as backend.  Examples are also provided (see `test.ikt`).


thebsdbox:
awesome!

PR #824, opened 2018-01-05
Check for changes to SSL cert and health monitor path.

craigyam:
A docker service update (or a docker service rm followed by a docker service create) which changes the certificate or health check path values may result in the load balancer configuration for the service having the configuration of the older service.

The ingress controller does not check if the certificate or health check path has changed between the Service Load Balancer configuration and the Docker Swarm Listener configuration. This results in the old configuration not getting updated to the new values.

### DIFF
--- a/pkg/controller/ingress/l4.go
+++ b/pkg/controller/ingress/l4.go
@@ -1,10 +1,16 @@
 package ingress
 
 import (
-	"fmt"
-
 	"github.com/docker/infrakit/pkg/controller/ingress/types"
 	"github.com/docker/infrakit/pkg/spi/loadbalancer"
+)
+
+type operation int
+
+const (
+	createOp operation = iota
+	changeOp
+	noop
 )
 
 // configureL4 configures a L4 loadbalancer with the desired routes and given options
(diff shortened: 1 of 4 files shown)


craigyam:
One thought I'd like to add...  Would there be any concerns with moving the removes before the creates?  I'm thinking of the case where the front end port was changed - the create would fail with dupe port (for IBM Cloud anyway) so the changed route wouldn't be added.  The remove would then remove that route and then another sync cycle would be needed before the route would be added.  Not sure how other providers handle that so wasn't sure if that would be a valid change.

pkg/controller/ingress/l4.go
```diff
@@ -110,13 +88,32 @@ func configureL4(elb loadbalancer.L4, desired []loadbalancer.Route, options type
 	return nil
 }
 
-func findRoutePort(routes []loadbalancer.Route, loadbalancerPort int,
-	protocol, loadbalancerProtocol loadbalancer.Protocol) (int, bool) {
-
-	for _, route := range routes {
-		if route.LoadBalancerPort == loadbalancerPort && route.Protocol == protocol && route.LoadBalancerProtocol == loadbalancerProtocol {
-			return route.Port, true
+func getRouteOperation(routes *[]loadbalancer.Route, desired loadbalancer.Route) operation {
```
chungers: Thanks.  This is much nicer than all that inline logic previously.

pkg/controller/ingress/l4_test.go
```diff
@@ -0,0 +1,228 @@
+package ingress
```
chungers: Thank you for adding tests!!

chungers:
> One thought I'd like to add...  Would there be any concerns with moving the removes before the creates?  I'm thinking of the case where the front end port was changed - the create would fail with dupe port (for IBM Cloud anyway) so the changed route wouldn't be added.  The remove would then remove that route and then another sync cycle would be needed before the route would be added.  Not sure how other providers handle that so wasn't sure if that would be a valid change.
  
This is a valid concern and the behavior you described would be typical.  Do you want to incorporate that in this PR or make a separate one?

craigyam:
Thanks David.  I will look at creating a new PR with the change to move the removesahead of the creates.

PR #826, opened 2018-01-05
Enrollment Options for parsing errors

kaufers:
The enrollment controller can encounter parsing errors in either the source instances or the enrolled instances.

A source instance failing to parse only affects the removal of currently enrolled instances (since we cannot determine the unique identifier of the source instance due to the parsing error and, hence, we cannot determine if the instance is currently enrolled). When a **source fails to parse** there are
two options:
- Enable the removal of one or more currently enrolled instances
- Disable the removal of any currently enrolled instances

**If removal is enabled**, then the source instance (that failed to parse) will be removed from the group (it is is currently enrolled).

When an **enrolled instance fails to parse**, this only affects the addition of source instances and there are two options:
- Enable the addition of one or more source instances
- Disable the addition of any source instances

**If addition is enabled**, the enrolled instance (that failed to parse) will be added again (if it is in the source instances).

This PR adds these `Options` to the enrollment controller configuration, the default is that removals and additions are **enabled** on parse errors.

### DIFF
--- a/pkg/controller/enrollment/controller.go
+++ b/pkg/controller/enrollment/controller.go
@@ -19,10 +19,10 @@ var (
 
 	// DefaultOptions return an Options with default values filled in.
 	DefaultOptions = enrollment.Options{
-		SyncInterval:         types.Duration(5 * time.Second),
-		DestroyOnTerminate:   false,
-		SourceParseErrOp:     enrollment.SourceParseErrorEnableDestroy,
-		EnrollmentParseErrOp: enrollment.EnrolledParseErrorEnableProvision,
+		SyncInterval:             types.Duration(5 * time.Second),
+		DestroyOnTerminate:       false,
+		SourceParseErrPolicy:     enrollment.SourceParseErrorEnableDestroy,
+		EnrollmentParseErrPolicy: enrollment.EnrolledParseErrorEnableProvision,
 	}
 )
 
--- a/pkg/controller/enrollment/controller.go
+++ b/pkg/controller/enrollment/controller.go
@@ -19,8 +19,10 @@ var (
 
 	// DefaultOptions return an Options with default values filled in.
 	DefaultOptions = enrollment.Options{
-		SyncInterval:       types.Duration(5 * time.Second),
-		DestroyOnTerminate: false,
+		SyncInterval:         types.Duration(5 * time.Second),
+		DestroyOnTerminate:   false,
+		SourceParseErrOp:     enrollment.SourceParseErrorEnableDestroy,
+		EnrollmentParseErrOp: enrollment.EnrolledParseErrorEnableProvision,
 	}
 )
 
--- a/pkg/controller/enrollment/enroller.go
+++ b/pkg/controller/enrollment/enroller.go
@@ -60,12 +60,37 @@ func newEnroller(scope scope.Scope, leader func() stack.Leadership, options enro
 		options: options,
 	}
 
-	if l.options.SyncInterval.Duration() == 0 {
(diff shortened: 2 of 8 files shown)


chungers:
I can see how this change might be controversial because this is largely a policy decision depending on what properties are used as key. Sometimes a zero value property may be the signal needed to remove something...

How about we add a field in Options to indicate the removal policy?  Like MissingFieldPolicy=remove  which can be set to make it more strict. Thus the enrollment controller is by default lenient?

kaufers:
@chungers PR has been updated to parse the desired operation from the `Options` object, see updated description for the supported configurations.

pkg/controller/enrollment/enroller.go
```diff
@@ -53,17 +53,20 @@ type enroller struct {
...
-
-	if l.options.SyncInterval.Duration() == 0 {
-		return nil, fmt.Errorf("sync interval cannot be 0")
+	if err := validateParseErrorOptions(l.options); err != nil {
+		return nil, err
+	}
+	// Note that the sync interval is valided in the constructor only since it cannot
+	// be modified after the enroller has started (ie, it cannot be changed on a spec
+	// update).
+	if l.options.SyncInterval.Duration() <= 0 {
```
chungers: How about we consolidate this duration check with the parse error options validation into one function and have all the logic there?

pkg/controller/enrollment/controller.go
```diff
@@ -19,8 +19,10 @@ var (
 
 	// DefaultOptions return an Options with default values filled in.
 	DefaultOptions = enrollment.Options{
-		SyncInterval:       types.Duration(5 * time.Second),
-		DestroyOnTerminate: false,
+		SyncInterval:         types.Duration(5 * time.Second),
+		DestroyOnTerminate:   false,
+		SourceParseErrOp:     enrollment.SourceParseErrorEnableDestroy,
```
chungers: Tiny nit:  how do you feel about naming this `SourceParseErrPolicy` and `EnrollmentParseErrPolicy`?

pkg/controller/enrollment/enroller.go
kaufers: @chungers see comment right above:
```
	// Note that the sync interval is valided in the constructor only since it cannot
	// be modified after the enroller has started (ie, it cannot be changed on a spec
	// update).
```
That's why I kept it out; thoughts?

pkg/controller/enrollment/controller.go
kaufers: Sure, that's trivial. I'll make that change.

pkg/controller/enrollment/enroller.go
chungers: Just a thought...

How about we do something like this in `enrollment/types/Options`

```
type PluginPhase int

const (
	PluginInit = iota
	PluginCommit
)

func (o Options) Validate(phase PluginPhase) error {
   if phase == PluginInit {
      // check the sync interval here too
   }
}
```
This validate function will tell me what's going on in terms of what is being checked during what lifecycle phases of the plugin.  We won't have validation code peppered throughout the enroller and having to figure out what's going on by understanding the call flow...
  
Also, with this we are adding some formalism to options validation and we might want to push this up to make `Validate(phase)` an interface that all the Options can implement.  Then we can move some of this code later out of the enroller.go and have the basic controller implementation enforce this validation...   

Thoughts?

kaufers: I like this idea; having a more formalized `Option` validation scheme sounds like a great idea.

PR #831, opened 2018-01-10
Cache terrform instances

kaufers:
This commit caches the output for the `DescribeInstances` command. The cache needs to be invalidated whenever one of the following operations is done:
- Provision
- Label
- Destroy

Also, after `terraform refresh` is executed, then the properties may have changed OR one or more input `tf.json` files may have been pruned. Therefore, the cache also needs to be invalidated after each refresh.

This commit also updates the locking to use a read/write lock so that multiple describe operations can run in parallel.

### DIFF
--- a/pkg/provider/terraform/instance/plugin.go
+++ b/pkg/provider/terraform/instance/plugin.go
@@ -1172,9 +1172,22 @@ func parseAttachTag(tf *TFormat) ([]string, error) {
 	return []string{}, nil
 }
 
+// External functions using during describe; broken out for testing
+type describeFns struct {
+	tfShow func(resTypes []TResourceType, propFilter []string) (map[TResourceType]map[TResourceName]TResourceProperties, error)
+}
+
 // DescribeInstances returns descriptions of all instances matching all of the provided tags.
 func (p *plugin) DescribeInstances(tags map[string]string, properties bool) ([]instance.Description, error) {
-	logger.Debug("DescribeInstances", "tags", tags, "V", debugV1)
+	fns := describeFns{
+		tfShow: p.doTerraformShow,
+	}
+	return p.doDescribeInstances(fns, tags, properties)
+}
+
+// doDescribeInstances returns descriptions of all instances matching all of the provided tags.
+func (p *plugin) doDescribeInstances(fns describeFns, tags map[string]string, properties bool) ([]instance.Description, error) {
+	logger.Debug("doDescribeInstances", "tags", tags, "V", debugV1)
 	// Acquire lock since we are reading all files and potentially running "terraform show"
 	p.fsLock.Lock()
 	defer p.fsLock.Unlock()
@@ -1190,10 +1203,10 @@ func (p *plugin) DescribeInstances(tags map[string]string, properties bool) ([]i
 		// TODO - not the most efficient, but here we assume we're usually just one vm type
 		for vmResourceType := range localSpecs {
 
-			if result, err := p.doTerraformShow([]TResourceType{vmResourceType}, nil); err == nil {
+			if result, err := fns.tfShow([]TResourceType{vmResourceType}, nil); err == nil {
 				terraformShowResult = result
 			} else {
-				logger.Warn("DescribeInstances", "terraform show error", err)
+				logger.Warn("doDescribeInstances", "terraform show error", err)
 				return nil, err
 			}
 		}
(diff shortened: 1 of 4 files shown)


pkg/provider/terraform/instance/plugin.go
```diff
@@ -1005,6 +1013,7 @@ func (p *plugin) Destroy(instID instance.ID, context instance.Context) error {
 	// Acquire Lock outside of recursive doDestroy function
 	p.fsLock.Lock()
 	defer p.fsLock.Unlock()
+	defer p.clearCachedInstances()
```
chungers: Is this really guarding the cache?  The clearing comes after the lock is released.   I'm  also not sure if multiple defers is stylistically...  how about 
```
defer func() {
    clearCache()
    lock.Unlock()
}()
```

chungers: Another thing. What about errors?  The defer will *always* clear cache even if we encounter an error.  Is that a good thing?

chungers: Heh https://medium.com/intuitionmachine/the-strange-loop-in-alphago-zeros-self-play-6e3274fcdd9f?source=linkShare-cc3cd841bbd8-1515523953

Forgot multiple defers are last in fist out (it's pushed onto a stack).  Anyway my point remains for readability:)

kaufers: @chungers the `defer` functions are pushed on a stack, so they are executed in reverse order: https://stackoverflow.com/questions/32541870/multiple-defers-vs-deferred-anonymous-function

But I do see the value is having a utility function for this, then it's consistent and the order is obviousl

kaufers: @chungers yeah, having it clear the cache on error probably isn't ideal. But, IMO, we'd want to error on the side or refreshing the cache vs attempting to differentiate between errors that invalidated the cache and those that didn't.

chungers: 1.  Yeah I prefer the closure approach for readability.  I wouldn't make a separate function just for that though since that utility function would look weird just releasing a lock.   

2. Ok that's reasonable.  I think it's a matter of policy and at this point I would emphasize correctness too.  So yes - let's clear the cache on all mutations.

pkg/provider/terraform/instance/plugin.go
```diff
@@ -1172,75 +1134,146 @@ func parseAttachTag(tf *TFormat) ([]string, error) {
...
-	p.fsLock.Lock()
-	defer p.fsLock.Unlock()
+	// Get the read lock and then check if the cache is populated
+	p.fsLock.RLock()
+
+	if p.cachedInstances == nil {
+		// Cache is not populated, now we need the write lock (after releasing the read lock) and
+		// then we can attempt to populate and use the cache
+		p.fsLock.RUnlock()
+		p.fsLock.Lock()
```
chungers: This code here that tries to refresh the cache by first obtaining the lock can  be moved into `refreshCachedInstances()` so that the write lock is acquired and released on entry/exit of that function.   Then the code here can just work with the read lock/unlock to guard the critical section.  OR you could have a function `isCacheNil()` that inside that function it's doing the acquire/release of read lock.  

This way we also have corresponding acquire/release of the locks, regardless of the lock type.  

What do you think?

kaufers: The problem is that the lock is not re-entrant; so we cannot acquire the write lock if we already have the read lock. I wanted move the write lock acquisition into a function but I found myself always needed to call that function _after_ I had the read lock.

chungers: I am thinking

```
func (p *plugin) isCacheNil() bool {
   p.lock.RLock()
   defer p.lock.Unlock()
   return p.cache == nil 
}

func (p *plugin) refreshCache() {
    p.lock.Lock() // write lock
    defer p.lock.Unlock()
    // check again after acquiring write lock
    if p.cache == nil {
        // do work
    }  
}

func (p *plugin) doDescribeInstances() {
   p.isCacheNil() { 
       p.refreshCache()        
   }
   // by now, we should have cached value in any case (nil or not nil)
   // there's no lock from this scope.

   p.lock.RLock()
   defer lock.RUnlock()
   // do filtering / computing a view of the cached value
}
```
 Will this work?

kaufers: @chungers Yes, thank you! See new commit that updated to this approach.

PR #832, opened 2018-01-11
Cleanup index error logging in ingress and enroller

kaufers:
Currently, the entire `instance.Description` object is logged, this includes all of the `instance.Properties` which is quite big (do to a long `userdata` boot script).

Instead of printing out the entire object, this commit updates the error message to print out the `instance.ID`, and the `instance.Tags` (which is enough to identity the instance).

The error messages pretty clearly show the reason for the parse error, there is no need to print out the entire `instance.Properties`. Some example errors are:
```
template: file://localhost/infrakit_files/:1:41: executing \"file://localhost/infrakit_files/\" at <$x.ipv4_address_priv...>: map has no entry for key \"ipv4_address_private\"
template: file://localhost/infrakit_files/:1:45: executing \"file://localhost/infrakit_files/\" at <$x.id>: map has no entry for key \"id\"
```

### DIFF
--- a/pkg/controller/enrollment/set.go
+++ b/pkg/controller/enrollment/set.go
@@ -21,7 +21,7 @@ func index(list instance.Descriptions, getKey keyFunc) (map[string]instance.Desc
 	for _, n := range list {
 		key, err := getKey(n)
 		if err != nil {
-			log.Error("cannot index entry", "instance.Description", n, "err", err)
+			log.Error("cannot index entry", "instance.ID", n.ID, "instance.tags", n.Tags, "err", err)
 			e = err
 			continue
 		}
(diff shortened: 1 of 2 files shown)


kaufers:
An example of the updated message:

```
EROR[01-11|17:28:14] cannot index entry                       module=controller/ingress instance.ID=instance-1515691603 instance.tags="map[infrakit-link-created:2018-01-11t17:26:43z infrakit.config_sha:fbjqqkmhtgi5noj5flvnre5o4cvcipq2 infrakit.group:workers name:instance-1515691603 swarm-id:l677dpzr9hghuxis2nk25soxc infrakit-link:a18mvwb0tda932wk infrakit-link-context:swarm::l677dpzr9hghuxis2nk25soxc::worker]" err="template: file://localhost/infrakit_files/:1:41: executing \"file://localhost/infrakit_files/\" at <$x.ipv4_address_priv...>: map has no entry for key \"ipv4_address_private\"" meta="{Identity:<nil> Name:d4ic-lb Tags:map[project:d4ic user:d4ic-user]}"
```

PR #833, opened 2018-01-12
Move terraform options into new terraform types

kaufers:
Creates a `types.go` that handles the definition of the terraform `Option`s and the associated functions.


pkg/provider/terraform/instance/types/types.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/docker/infrakit/pkg/types"
+	"github.com/docker/machine/libmachine/log"
+)
+
+var (
+	logger = logutil.New("module", "provider/terraform/instance/types")
+)
+
+// ImportResourceOptions defines a resource to import
+type ImportResourceOptions struct {
```
chungers: Why not have this as a field in the overall `Options`?

kaufers: Ahh, good idea; I wasn't even thinking of that when I pulled it from the current implementation.

kaufers: Actually, it's on line 52 -- is that not what you meant?

kaufers: So I think that part of the confusion here is that I left some of the "type" definitions in the `plugin.go`, mainly because they are everywhere in the code (and the test code). Things like:
```
// TResourceType is the type name of the resource: e.g. ibmcloud_infra_virtual_guest
type TResourceType string

// TResourceName is the name of the resource e.g. host1
type TResourceName string

// TResourceProperties is a dictionary for the resource
type TResourceProperties map[string]interface{}

// TFormat is the on-disk file format for the instance-xxxx.json.  This supports multiple resources.
type TFormat struct {
	// Resource matches the resource structure of the tf.json resource section
	Resource map[TResourceType]map[TResourceName]TResourceProperties `json:"resource"`
}
```

This should really be in this new `types.go` file, right? While that will be painful, if that's the intent of this type of file then I'm fine with doing the move. Thoughts?

chungers: Ok I see it now... This looks good.

Would you be open to calling this struct just `Resource`?  Essentially it's a model struct that captures the `Type`, `Name`, and `ID` of a resource to be imported.    So perhaps for brevity, just:

```
type Resource struct {
    Type string
    Name string
    ID string
    ExcludedPropIDs []string
}

// and in the Options
type Options struct {
   ImportResources []Resource
}
```
Other than the naming nitpick... everything looks good.  Thanks for creating a `types.go`!

kaufers: @chungers done, renamed the struct in a new commit; thanks for the review!

PR #835, opened 2018-01-16
Update swarm ID tag to const in flavor SPI

kaufers:
Similar to https://github.com/docker/infrakit/pull/810

### DIFF
--- a/pkg/spi/flavor/spi.go
+++ b/pkg/spi/flavor/spi.go
@@ -8,8 +8,8 @@ import (
 )
 
 const (
-	// SwarmIDTag is the name of the tag that contains the swarm ID
-	SwarmIDTag = "swarm-id"
+	// ClusterIDTag is the name of the tag that contains unique ID for the cluster
+	ClusterIDTag = "infrakit.cluster.id"
 )
 
 // InterfaceSpec is the current name and version of the Flavor API.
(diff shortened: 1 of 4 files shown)


pkg/spi/flavor/spi.go
```diff
@@ -7,6 +7,11 @@ import (
 	"github.com/docker/infrakit/pkg/types"
 )
 
+const (
+	// SwarmIDTag is the name of the tag that contains the swarm ID
+	SwarmIDTag = "swarm-id"
```
chungers: Would you mind renaming this to `ClusterIDTag`?  

At the level of SPI, which is agnostic to *what* flavor it is, seeing a reference to Swarm feels strange.    I think it's totally reasonable to have something generic like a "cluster" or "stack" ID.  Indeed we already have the concept of stack elsewhere in the code base (e.g. the manager manages a "stack")

kaufers: @chungers sure, I'm fine with the constant var name. How about the value of the tag? Should it also change from `swarm-id` to `cluster-id`? Or, to be more consistent with other tags, should it be `swarm.id` or `cluster.id`?

chungers: How about `infrakit.cluster.id` ?

I made a similar comment on #836  -- I think it'd be useful to follow a convention for tags that are created by infrakit.

kaufers: Sure, I like that, it'll be:
```
ClusterIDTag = "infrakit.cluster.id"
```
Thanks for the review!

PR #836, opened 2018-01-17
Update terraform instance name tag

kaufers:
Updates the tag key from `Name` to `infrakit.infrakit.name` so that it aligns with the other tags.

### DIFF
--- a/pkg/provider/terraform/instance/plugin.go
+++ b/pkg/provider/terraform/instance/plugin.go
@@ -52,6 +52,9 @@ const (
 
 	// scopeGlobal is the scope key for global resources
 	scopeGlobal = "global"
+
+	// NameTag is the name of the tag that contains the instance name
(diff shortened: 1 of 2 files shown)


pkg/provider/terraform/instance/plugin.go
```diff
@@ -52,6 +52,9 @@ const (
 
 	// scopeGlobal is the scope key for global resources
 	scopeGlobal = "global"
+
+	// NameTag is the name of the tag that contains the instance name
+	NameTag = "instance.name"
```
chungers: Can we prefix this with `infrakit.` so that when users look at the resources and their tags they can tell this is an infrakit managed resource?

kaufers: Sure, it'll be:
```
NameTag = "infrakit.instance.name"
```
Thanks for the review!

PR #844, opened 2018-01-29
Pace quorum group rolling updates

kaufers:
Currently, the `Allocation.Size` value is used to determine group size; this only works for `scaler` groups, for `quorum` groups we need to use the `Allocation.LogicalIDs` value.

### DIFF
--- a/pkg/plugin/group/rollingupdate.go
+++ b/pkg/plugin/group/rollingupdate.go
@@ -67,7 +67,6 @@ type rollingupdate struct {
 	scaled       Scaled
 	updatingFrom groupSettings
 	updatingTo   groupSettings
-	desiredSize  int
 	stop         chan bool
 }
 
@@ -123,13 +122,13 @@ func (r *rollingupdate) waitUntilQuiesced(pollInterval time.Duration, expectedNe
 				// the group, and/or documenting the expectations for plugin implementations.
 				switch r.scaled.Health(inst) {
 				case flavor.Healthy:
-					log.Info("waitUntilQuiesced", "msg", fmt.Sprintf("Node %s is Healthy", inst.ID))
+					log.Info("waitUntilQuiesced", "health", "heathy", "nodeID", inst.ID)
 					numHealthy++
 				case flavor.Unhealthy:
-					log.Warn("waitUntilQuiesced", "msg", fmt.Sprintf("Node %s is Unhealthy", inst.ID))
+					log.Error("waitUntilQuiesced", "health", "unheathy", "nodeID", inst.ID)
 					return fmt.Errorf("Instance %s is unhealthy", inst.ID)
 				case flavor.Unknown:
-					log.Info("waitUntilQuiesced", "msg", fmt.Sprintf("Node %s has Unknown health", inst.ID))
+					log.Info("waitUntilQuiesced", "health", "unknown", "nodeID", inst.ID)
 				}
 			}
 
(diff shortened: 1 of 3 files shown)


pkg/plugin/group/rollingupdate.go
```diff
@@ -120,17 +123,27 @@ func (r *rollingupdate) waitUntilQuiesced(pollInterval time.Duration, expectedNe
 				// the group, and/or documenting the expectations for plugin implementations.
 				switch r.scaled.Health(inst) {
 				case flavor.Healthy:
+					log.Info("waitUntilQuiesced", "msg", fmt.Sprintf("Node %s is Healthy", inst.ID))
```
chungers: The convention for logging is something like
```
log.Info("waitUntilQuiesced - node healthy", "nodeID", inst.ID)
```

Also, I don't know if we need to have a logging statement for each case...  maybe just one statement before the switch?

```
health:= r.scaled.Health(inst)
log.Info("waitUntilQuiesced", "nodeID", inst.ID, "health", health)
switch health {
...
```

pkg/plugin/group/rollingupdate.go
```diff
@@ -149,37 +162,37 @@ func (r *rollingupdate) Run(pollInterval time.Duration) error {
...
 	desired, _ := desiredAndUndesiredInstances(instances, r.updatingTo)
 	expectedNewInstances := len(desired)
+	log.Info("RollingUpdate-Run", "expectedNewInstances", expectedNewInstances)
 
 	for {
-		err := r.waitUntilQuiesced(
-			pollInterval,
-			minInt(expectedNewInstances, int(r.updatingTo.config.Allocation.Size)))
+		// Wait until any new nodes are healthy
+		err := r.waitUntilQuiesced(pollInterval, minInt(expectedNewInstances, r.desiredSize))
```
chungers: Since `r.desiredSize` was initialized to `int(config.Allocation.Size)`, isn't this code exactly the same as before?

pkg/plugin/group/scaler.go
```diff
@@ -118,6 +118,7 @@ func (s *scaler) PlanUpdate(scaled Scaled, settings groupSettings, newSettings g
 		scaled:       scaled,
 		updatingFrom: settings,
 		updatingTo:   newSettings,
+		desiredSize:  int(settings.config.Allocation.Size),
```
chungers: if we initialize this value here...  does this ever get updated when we change the count (Allocation.Size) of a group?

pkg/plugin/group/rollingupdate.go
chungers: `r.desiredSize` is referenced here only to determine the min count to wait on.  `r.desiredSize` is initialized to be either `len(config.Allocations.LogicalIDs)` or `int(config.Allocations.Size)`...    Do you think it'd be more direct to just determine the `desiredSize` here... something like

```
desiredSize := len(r.updatingTo.config.Allocation.LogicalIDs)
if desiredSize == 0 {
   desiredSize = int(r.updatingTo.config.Allocation.Size)
}
err := r.waitingUntilQuiesced(pollInterval, minInt(expectedNewInstances, desiredSize))
...
```

Then we don't even need the extra field in `r`?   seems like this isn't something the user can set anyway..

pkg/plugin/group/rollingupdate.go
kaufers: I can update the format, my motivation for being more verbose in the logging was because `health` is just an `int`.  IMO the log messages are more obvious when they have messages like `Unknown`/`Healthy`/`Unhealthy` vs `0`/`1`/`2`. Thoughts?

pkg/plugin/group/rollingupdate.go
kaufers: Yeah, this makes sense and then we don't need to have the struct value.

pkg/plugin/group/rollingupdate.go
chungers: Ah ok makes sense.   Then we probably don't need the `fmt.Sprintf()` then... just something like
```
log.Info("waitUntilQuiesed", "health", "heathy", "nodeID", inst.ID)
```
in each one of the `case`s.

pkg/plugin/group/rollingupdate.go
chungers: I think computing it there would make the logic easier to follow and would address the question I had in `scaler.go` - about mutating that field when the group size is changed...

Issue #846, opened 2018-02-01
Swarm node garbage collection


kaufers:
See "Garbage Collection" section in https://github.com/docker/infrakit/issues/838#issuecomment-360625772

There are 3 different scenarios that we want to handle. To illustrate these, assume that we have 1 group plugin (`SWARM-GRP-READY` that returns all `ready` swarm nodes) and 2 instance plugins (`SWARM-INST-ALL` that returns all swarm nodes and `VM-INST-ALL` which is the generic instance plugin that is wired to the group controller).

## Scenario 1: Orphaned swarm node
In this case, assume that node `n2` was the old leader and it has been demoted but it could not be removed from the swarm. The `VM-INST-ALL` plugin does not have any entry for it since it was `Destroy`d (the `flavor.Drain` was also completed in that flow).
```
SWARM-GRP-READY   SWARM-INST-ALL   VM-INST-ALL
n1-link1          n1 (ready)       n1-link1
                  n2 (down)                   
n3-link3          n3 (ready)       n3-link3
```
In this case, we wire the enrollment controller to use `SWARM-GRP-READY` and the `SWARM-INST-ALL`; since `n2` is missing from the group then the `SWARM-INST-ALL.Destroy` is executed remove the orphan from the swarm.

**NOTE: that this presents the same as scenario 3 (from the perspective of `SWARM-INST-ALL`)**

## Scenario 2: Node fails to join
In this case, assume that node `n2` was `VM-INST-ALL.Provision`d but it never joined.
```
SWARM-GRP-READY   SWARM-INST-ALL   VM-INST-ALL
n1-link1          n1 (ready)       n1-link1
                                   n2-link2
n3-link3          n3 (ready)       n3-link3
```
In this case, we wire the enrollment controller to use `SWARM-GRP-READY` and the `VM-INST-ALL`; since `n2` is missing from the group then it is removed from `VM-INST-ALL`. The group controller will then detect that the group is not at the desired size and issue another `VM-INST-ALL.Provision`.

## Scenario 3: Node goes offline
In this case, assume that node `n2` was `Provision`d and joins the cluster and then it goes offline.
```
SWARM-GRP-READY   SWARM-INST-ALL   VM-INST-ALL
n1-link1          n1 (ready)       n1-link1
                  n2 (down)        n2-link2
n3-link3          n3 (ready)       n3-link3
```
[...]

PR #853, opened 2018-02-06
Use fsm to model the workflow of garbage collections

chungers:
This PR implements a simple POC for #846 using the pkg/fsm state machine models. 
The POC is implemented as more tests for the fsm package

   + Some data races have been fixed by introducing locks
   + The POC that models the garbage collection problem is in `usage_gc_test.go`.  

Some terminology:  `Index` type is for defining possible states while `Signal` are signals that a fsm will react to.   For the purpose of managing swarm nodes, we consider a swarm node to have two components that are joined via a link label.    For linking a vm instance to a running Docker engine, there must be a tag of value `K` for the vm instance and a Docker engine label of value `K`.  Once this association is established, we no longer need to query the actual Docker engine (via TLS or via SSH) for engine identity and our fsm models this using the states:
```
const (
	start                    Index = iota
	matched_instance               // has vm information, waiting to match to docker_node
	matched_docker_node            // has docker_node information, waiting to match to vm
	swarm_node                     // has matching docker_node and vm information
	swarm_node_ready               // ready as swarm node
	swarm_node_down                // unavailable as swarm node
	pending_instance_destroy       // vm needs to be removed_instance (instance destroy)
	removed_instance               // instance is deleted
	done                           // terminal
)
```
with signals that we can implement:
```
const (
	docker_node_ready Signal = iota
	docker_node_down
	docker_node_gone
	instance_ok
	instance_gone
	timeout
	reap
)
```
Note that we can determine when a Docker Node or when an instance is `gone` by computing the difference of the sampled sets of nodes/instances.  Please see `TestSwarmEntities` test for the four cases listed in #846 
  1. Orphaned node
  2. Node fails to join
  3. Node goes offline
  4. Rouge node

Here is an example state definition:
```
		State{
			Index: matched_docker_node,
			TTL:   Expiry{wait_describe_instances, instance_gone},
			Transitions: map[Signal]Index{
				instance_ok:      swarm_node,
				instance_gone:    removed_instance,
				docker_node_gone: removed_instance, // could be docker rm'd out of band
			},
			Actions: map[Signal]Action{
				instance_gone: dockerNodeRm.do,
			},
		},
```
In this example, the state `matched_docker_node` denotes the state a particular "node" is in when there's a `Docker node ls` entry.   In the `Transitions` map, a signal `instance_ok` will transition the state to `swarm_node`, while `instance_gone` will transition to `removed_instance`.  In this state we can define a TTL that says we can only be in this state for `wait_describe_instances` * ticks, where ticks is a polling interval the fsm engine is initialized with.   When the deadline is exceeded, the signal `instance_gone` is raised and an action called `dockerNodeRm.do` is performed, as specified in the `Actions` map.

@kaufers 


chungers:
merging this because regardless of which implementation design we take, this PR adds at a minimum more tests to the fsm package.

PR #856, opened 2018-02-07
Issue second group commit on manager leadership change

kaufers:
Currently, the groups are committed on manager leadership change; this initial commit simply starts the group polling. A second commit is needed to complete and in progress updates (for example, completing a rolling update on a quorum group when leadership changes to a new manager).

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -438,27 +438,10 @@ func (m *manager) onLostLeadership() error {
 	return m.doFreeAll(config)
 }
 
-func (m *manager) doCommit() error {
-
-	defer m.metadataChanged()
-
-	// load the config
-	config := globalSpec{}
-
-	// load the latest version -- assumption here is that it's been persisted already.
-	err := config.load(m.Options.SpecStore)
-	if err != nil {
-		return err
-	}
-
-	log.Info("Committing. Loaded snapshot.", "err", err)
-	if err != nil {
-		return err
-	}
-	return m.doCommitAll(config)
-}
-
 func (m *manager) doCommitAll(config globalSpec) error {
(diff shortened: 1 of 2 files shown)


pkg/manager/manager.go
```diff
@@ -475,7 +458,8 @@ func (m *manager) doCommitAll(config globalSpec) error {
 				log.Error("Cannot commit group", "spec", spec, "err", err)
 			}
 			return true, err
-		})
+		},
+		true)
```
chungers: Can you make this `true` a parameter of this `doCommitAll()` function?  Then in the code near `onAssumeLeadership` we pass this `true` explicitly and then in a regular `Commit` we pass false?  Then we can add a comment on the meaning of issuing two commits on leadership failover.  This should make the logic easier to follow.  Thoughts?

kaufers: Well, `doCommitAll` is only invoked from  `loadAndCommitSpecs` which is only invoked from `onAssumeLeadership`. So, if we want to make it more _explicit_ on `onAssumeLeadership` then we need to pass this new `bool` through. But, since a "commit" only happens when leadership is assumed then it's this already intuitive?

I'm all for changing the flow (since it took me a long time to understand this code) but I don't know if this proposed change will do it.

The motivation behind this additional param was to do the 2nd group commit after all of the groups were originally committed. In theory, we could instead just call commit 2 times in a row per group. Thoughts?

chungers: Yup I agree with you.  Can you add a comment at the `true` to mention it's needed to continue updating in the event of failover?

kaufers: Yep, see new commit.

PR #859, opened 2018-02-09
Flavor updates

kaufers:
- Enhance the definition of a `Healthy` node to be a node that is in a `ready` state and, if it is a manger, that it is `reachable`
- Update manager `Drain` to only demote the node
- Change rolling update quiesced loop to handle Unhealthy nodes

### DIFF
--- a/pkg/plugin/flavor/swarm/flavor.go
+++ b/pkg/plugin/flavor/swarm/flavor.go
@@ -218,6 +218,28 @@ func (s *baseFlavor) Healthy(flavorProperties *types.Any, inst instance.Descript
 		return flavor.Unknown, nil
 
 	case len(nodes) == 1:
+		if nodes[0].Status.State != swarm.NodeStateReady {
+			log.Warn("Node is not ready",
+				"id", nodes[0].ID,
+				"hostname", nodes[0].Description.Hostname,
+				"state", nodes[0].Status.State)
+			return flavor.Unhealthy, nil
+		}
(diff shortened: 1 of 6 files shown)


pkg/plugin/flavor/swarm/flavor.go
```diff
@@ -218,6 +218,28 @@ func (s *baseFlavor) Healthy(flavorProperties *types.Any, inst instance.Descript
...
 
 	case len(nodes) == 1:
+		if nodes[0].Status.State != swarm.NodeStateReady {
+			log.Warn("Node is not ready",
+				"id", nodes[0].ID,
+				"hostname", nodes[0].Description.Hostname,
+				"state", nodes[0].Status.State)
+			return flavor.Unhealthy, nil
+		}
+		if nodes[0].Spec.Role == swarm.NodeRoleManager {
```
chungers: Since this function is used for both manager and workers, can we make a `switch role` and check for the cases of `Manager` and `Worker` and return healthy/unhealthy accordingly?  I see only the manager case here..

kaufers: All of the code in this conditional is only applicable to a manager node (since it's checking `ManagerStatus.Reachability`). The "common" check is for the `ready` state (which is above). Do you still want a switch if only the manager role is going to have a case statement?

chungers: Ok I see your point.  It's saying that the node has to be minimally 'Ready' and for managers another level of checks will be run to determine the health.

Let's keep it as-is then.

PR #861, opened 2018-02-09
Wait for node to be healthy for a period of time during update

kaufers:
Adds a new `Updating` struct on the `group.Spec`:
```
type Updating struct {
	Duration types.Duration
	Count    int
}
```
This defines how long a node needs to be healthy on a rolling update before the update is applied to the next node in the group.

### DIFF
--- a/pkg/controller/group/rollingupdate.go
+++ b/pkg/controller/group/rollingupdate.go
@@ -78,11 +78,19 @@ func (r rollingupdate) Explain() string {
 	return r.desc
 }
 
-func (r *rollingupdate) waitUntilQuiesced(pollInterval time.Duration, expectedNewInstances int) error {
+// Tracks the progress of the update
+type updatingCount struct {
+	healthyTs    *time.Time
+	healthyCount *int
+}
+
+func (r *rollingupdate) waitUntilQuiesced(pollInterval time.Duration, updating group_types.Updating, expectedNewInstances int) error {
 	// Block until the expected number of instances in the desired state are ready.  Updates are unconcerned with
 	// the health of instances in the undesired state.  This allows a user to dig out of a hole where the original
 	// state of the group is bad, and instances are not reporting as healthy.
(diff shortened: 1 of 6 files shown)


pkg/controller/group/rollingupdate.go
```diff
@@ -154,10 +166,54 @@ func (r *rollingupdate) waitUntilQuiesced(pollInterval time.Duration, expectedNe
 	}
 }
 
+// exceedsHealthyThreshold returns true for the Updating threshold is exceeded
+func exceedsHealthyThreshold(updating group_types.Updating, expectedNewInstances int, counts *updatingCount) bool {
```
chungers: Any reason why `counts` is a pointer?  There are no nil checks in this function.  If it's never nil then maybe we don't need pointer...

kaufers: We want to update the `struct` members, if we don't pass a pointer then a copy is passed and it's not mutated.

I originally had this logic embedded into the `waitUntilQuiesced` function but it got too nested and harder to test.

chungers: Ah I see this is meant to update some state that is used in the caller...

Another option is to have a new `updatingCount` value as the return value, along with the bool:

```
func exceedsHealthyThreshold(Updating, int, updatingCount) (updatingCount, bool)
```
So the return value is the next state.  
Sorry I am being pedantic -- I generally prefer functions that doesn't have side effects -- taking in input, returning output and not mutating something inside the function (unless the function has a receiver now -- which you can make this a "method" on the `updatingCount` struct....    

Don't know if this makes sense to you.  Your call :)

kaufers: Ahh, that's a great point. I'll change the function to have a receiver:
```
func (counts *updatingCount) exceedsHealthyThreshold(updating group_types.Updating, expectedNewInstances int) bool {
```

PR #864, opened 2018-02-12
Destroy current VM last in group destroy

kaufers:
When destroying a group, if one of the members is "self" then we should destroy it last. We do not want to destroy the VM that is running the instance plugin before it has completed all of the VM terminations.

### DIFF
--- a/pkg/controller/group/group.go
+++ b/pkg/controller/group/group.go
@@ -3,6 +3,7 @@ package group
 import (
 	"errors"
 	"fmt"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -181,18 +182,10 @@ func (p *gController) DestroyGroup(gid group.ID) error {
 		if err != nil {
 			return err
 		}
-		// Destroy self last
-		selfIndex := -1
-		for index, desc := range descriptions {
-			if isSelf(desc, context.settings) {
-				selfIndex = index
-				continue
-			}
+		// Ensure that the current node is last
+		sort.Sort(sortByID{list: descriptions, settings: &context.settings})
+		for _, desc := range descriptions {
 			context.scaled.Destroy(desc, instance.Termination)
-		}
-		if selfIndex != -1 {
-			log.Info("DestroyGroup, destroying self last", "self", *context.settings.options.Self)
-			context.scaled.Destroy(descriptions[selfIndex], instance.Termination)
 		}
 	}
 
--- a/pkg/controller/group/group.go
+++ b/pkg/controller/group/group.go
@@ -181,9 +181,18 @@ func (p *gController) DestroyGroup(gid group.ID) error {
 		if err != nil {
 			return err
 		}
-
-		for _, desc := range descriptions {
+		// Destroy self last
+		selfIndex := -1
+		for index, desc := range descriptions {
+			if isSelf(desc, context.settings) {
+				selfIndex = index
+				continue
+			}
 			context.scaled.Destroy(desc, instance.Termination)
+		}
+		if selfIndex != -1 {
+			log.Info("DestroyGroup, destroying self last", "self", *context.settings.options.Self)
+			context.scaled.Destroy(descriptions[selfIndex], instance.Termination)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/group/group.go
```diff
@@ -181,10 +181,19 @@ func (p *gController) DestroyGroup(gid group.ID) error {
 		if err != nil {
 			return err
 		}
-
-		for _, desc := range descriptions {
+		// Destroy self last
```
chungers: The code that orders the slice of instances to destroy will make sure the self node is always the last entry in the slice https://github.com/docker/infrakit/blob/master/pkg/controller/group/state.go#L129

So is this still necessary?   If anything we should remove the policy checks and just make sure the self node is always last from the sort.

chungers:
In addition to my comment on the sort order of the instances to destroy, the bigger question is whether the self vm should terminate itself.  

If we have a swarm demote then the self node should be terminated by the new leader, right?  I have concerns about the state on disk of the terraform plugins when the host goes down in the middle of a destroy.

pkg/controller/group/group.go
kaufers: So that sorting is **only** done in the **rolling update** flow: https://github.com/docker/infrakit/blob/master/pkg/controller/group/rollingupdate.go#L252

The **group destroy** flow being updated in this PR.

What do you think about updating the generic `scaled.List` to apply the same sort? https://github.com/docker/infrakit/blob/master/pkg/controller/group/scaled.go#L137

The benefit of this is that the current leader will always be returned last for any `List` operation and no addition work is needed to ensure that "self" is done last.

kaufers:
> the bigger question is whether the self vm should terminate itself.

Yeah, so this gets interesting. We need to allow the manager flavor `Drain` in the destroy flow. But should we allow the instance plugin to actually `Destroy` itself? Or do we prevent the invocation of the instance plugin's `Destroy` API?

> If we have a swarm demote then the self node should be terminated by the new leader, right? I have concerns about the state on disk of the terraform plugins when the host goes down in the middle of a destroy.

Actually, today, the instance plugin's `Destroy` is called. The tf implementation removes the associated `tf.json` file from disk. However, `terraform apply`, is never invoked because we only invoke that on the leader (and it's no longer a leader).

Then, when the new leader is elected, the tf instance's `DescribeInstances` API only returns 2 (out of 3) managers (since it's based on the `tf.json` data and the "old" leader's `tf.json` file was removed) so then the group size constraint is not met and another manager is spun up. So, in reality, it's not completing the rolling update and it's just ensuring that the size constraint is satisfied.

With this background, seems like we should allow the flavor `Drain` but prevent the instance `Destroy`? Thoughts @chungers

chungers:
> If we have a swarm demote then the self node should be terminated by the new leader, right? I have concerns about the state on disk of the terraform plugins when the host goes down in the middle of a destroy.
Actually, today, the instance plugin's Destroy is called. The tf implementation removes the associated tf.json file from disk. However, terraform apply, is never invoked because we only invoke that on the leader (and it's no longer a leader).
>
> Then, when the new leader is elected, the tf instance's DescribeInstances API only returns 2 (out of 3) managers (since it's based on the tf.json data and the "old" leader's tf.json file was removed) so then the group size constraint is not met and another manager is spun up. So, in reality, it's not completing the rolling update and it's just ensuring that the size constraint is satisfied.
>
> With this background, seems like we should allow the flavor Drain but prevent the instance Destroy? Thoughts @chungers

That's what I am thinking...   here's some logic to check for this:  https://github.com/docker/infrakit/blob/master/pkg/controller/group/rollingupdate.go#L40

I think the extraneous policy flags obfuscates the intent here....   Perhaps we should make it a rule to *never* call instance Destroy on self, instead of all this confusing policy check...

If the self vm nevers calls instance.Destroy, then we'd never modify the tf.json files on disk.  Instead, we just let leadership transfer to a new node that was just updated...  That new leader will have a clean view of the infrastructure -- it will see that one instance needs an update in the manager group and go through a normal sequence of Destroy and reprovision and re-join as manager.  It can make this decision purely looking at the hash tag of the instance and not based on any information from the swarm...  then it removes a tf.json file and can call a tf apply.  There shouldn't be any races going on.    What do you think?

kaufers:
I've opened up https://github.com/docker/infrakit/issues/867 to handle the changes to the rolling update policies.

PR #872, opened 2018-02-16
Update go build constraints

kaufers:
- Break out providers into unique files so that individual ones can be included in a build
- Group all providers into a common `provider.go` (with the exception of `libvirt` and `ucs`)
- Group all builtin plugins into a common `builtin.go`


Vanuan:
Why is libvirt excluded from builtin providers?

Vanuan:
Found it: #713 #718

